PR #9, opened 2023-06-23
Add MCT, DDT, DREAL, DEAL to org list

streino:
@abulte Feel free to modify the list if this is too much for a start.


abulte (approved):
🎉 such orgz

PR #10, opened 2023-06-27
Paginate datasets and organizations

abulte:
Due to the new config file #9, it became necessary to paginate organizations and datasets.

This PR also introduces a clean(ish) organizations list page with custom cards (vue-dsfr ones too limited) and a (dumb) loading state overlay since we're now making a lot more requests.

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -3,8 +3,8 @@
 # data.gouv.fr API base URL
 datagouvfr_api_url: https://www.data.gouv.fr/api
 
-frontend_setttings:
-  organizations_list_page_size: 8
+# display settings
+organizations_list_page_size: 8
 
 # list of organisations' ids that should be handled by the portal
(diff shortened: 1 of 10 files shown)


config.yaml
```diff
@@ -3,6 +3,9 @@
 # data.gouv.fr API base URL
 datagouvfr_api_url: https://www.data.gouv.fr/api
 
+frontend_setttings:
```
streino, line 6: typo in settings (I would just remove the term, everything in config.yaml is a setting anyway)

src/store/index.js
```diff
@@ -45,26 +80,68 @@ export const useOrganizationStore = defineStore("organization", {
...
+          href: "#",
+          title: `Page ${page}`,
+        }
+      })
+    },
+    /**
+     * Get orgs list for a given page from store
+     *
+     * @param {number} page
+     * @returns
```
streino, line 105: Missing type

PR #11, opened 2023-06-27
Add dataset cards

abulte:


streino (approved):
Misc comment:
On the org detail page, instead of putting the logo on each dataset card, I'd put it a on top of the page along with the org description.
Makes the card a bit more complex though, so not a priority at all.

abulte:
@streino Agreed. We'll need to iterate on the design later anyway. ATM I'm just bootstrapping stuff with an acceptable look.

PR #28, opened 2023-06-28
Dataset detail page

abulte:
Adds a dataset detail page.

A pretty big refactoring was needed at the API wrapper level since working with `ref()`s nested at different level (fetching a single dataset from the API vs extracting it from organization data) was becoming a nightmare.

The new async/await logic is cleaner IMHO, yet a bit more verbose in components.

New logic:
- stores have async `load*()` methods that trigger API calls that can be awaited, to be called on `onMounted()` and when something changes, eg the page being viewed
- stores have dynamic synchronous getters `get*()` on the results of the API calls and parameters such as page

### DIFF
--- a/src/router/index.js
+++ b/src/router/index.js
@@ -2,6 +2,7 @@ import { createRouter, createWebHistory } from "vue-router"
 import HomeView from "../views/HomeView.vue"
 import OrganizationsListView from "../views/organizations/OrganizationsListView.vue"
 import OrganizationDetailView from "../views/organizations/OrganizationDetailView.vue"
+import DatasetDetailView from "../views/datasets/DatasetDetailView.vue"
 
 const router = createRouter({
   history: createWebHistory(import.meta.env.BASE_URL),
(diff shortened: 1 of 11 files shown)


src/router/index.js
```diff
@@ -30,14 +31,18 @@ const router = createRouter({
...
-      // TODO:
-      component: () => import("../views/AboutView.vue"),
+      component: DatasetDetailView,
     },
     {
       path: "/about",
       name: "about",
       component: () => import("../views/AboutView.vue"),
     },
+    {
```
streino, line 41:
```
[plugin:vite:import-analysis] Failed to resolve import "../playground/Component.vue" from "src/router/index.js". Does the file exist?

.../ecospheres-front/src/router/index.js:44:31

42 |        path: "/playground",
43 |        name: "playground",
44 |        component: () => import("../playground/Component.vue"),
   |                                ^
45 |      },
46 |    ],
```

abulte: Oups sorry, I left my debug route. Fixed.

streino (approved):

PR #29, opened 2023-06-29
Add build index script for search

abulte:
This is quick and dirty script building an index that could be used by [lunr](https://lunrjs.com) for a frontend search.

The output index size is 82Mo with our current config, only with the `title` and `description` fields. Even with gzip, I don't think it will be small enough to be usable :-/

This PR is just for reference.


streino (commented):
Seems also this would only work for queries, not facets.

PR #30, opened 2023-06-29
External search prototype

abulte:
Designed to work with https://github.com/abulte/ecospheres-search-index, based on Meilisearch.

Not sure if we'll go with this but I guess it shows it can be quite simple, at least for basic features.

### DIFF
--- a/index.html
+++ b/index.html
@@ -4,7 +4,7 @@
     <meta charset="UTF-8">
     <link rel="icon" href="/favicon.ico">
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
-    <title>Vite App</title>
+    <title>Ecosphères</title>
   </head>
   <body>
     <div id="app"></div>
--- a/src/views/datasets/DatasetsListView.vue
+++ b/src/views/datasets/DatasetsListView.vue
@@ -1,0 +1,38 @@
+<script setup>
+import { computed, onMounted, watch } from "vue"
+import { useRoute } from "vue-router"
+import Card from "../../components/Card.vue"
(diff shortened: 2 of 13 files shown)


streino (commented):
Meilisearch looks cool!
Except for complex geo search (bbox, polygon), looks like it has everything we'd need.

index.html
```diff
@@ -4,7 +4,7 @@
     <meta charset="UTF-8">
     <link rel="icon" href="/favicon.ico">
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
-    <title>Vite App</title>
+    <title>Ecosphères</title>
```
streino, line 7: Now it looks real ;)

src/views/datasets/DatasetsListView.vue
```diff
@@ -0,0 +1,40 @@
...
+        type="dataset"
+        :alt-img="d.title"
+        :link="`/datasets/${d.slug}`"
+        :title="d.title"
+        :description="d.description"
+        :img="d.organization.logo"
+      />
+    </div>
+  </div>
+  <DsfrPagination v-if="pages.length" :current-page="currentPage - 1" :pages="pages" @update:current-page="onUpdatePage" />
```
streino, line 39: This doesn't do anything at this point, right?

streino (approved):

src/views/datasets/DatasetsListView.vue
abulte: Right. I'll remove it since all variables are not initialized.

PR #32, opened 2023-07-05
Add discussions for a dataset

abulte:


src/store/DiscussionStore.js
```diff
@@ -0,0 +1,64 @@
+import { defineStore } from "pinia"
```
streino, line 1: I wonder if the various stores could be refactored to share more code?
Not a priority, maybe I'll give it a shot later if you don't get to it.

src/views/datasets/DatasetDetailView.vue
```diff
@@ -72,4 +94,41 @@ onMounted(() => {
       :img="r.organization?.logo || r.owner.avatar"
     />
   </div>
+
+  <h2 class="fr-mt-4w">Discussions</h2>
+  <div v-if="!discussions.data?.length">Pas de discussion pour ce jeu de données.</div>
+
+  <DsfrAccordionsGroup>
```
streino, line 101: I'd find it easier to read if it was all expanded by default, with the possibility to collapse what we want. But that's more of a cosmetic change, not high priority.

streino (approved):

src/views/datasets/DatasetDetailView.vue
abulte: Unfortunately the accordion component does not support that: https://vue-dsfr.netlify.app/?path=/docs/composants-dsfraccordionsgroup--docs — only one item can be expanded at a time. We'll need to change the component used if we want a different look.

src/store/DiscussionStore.js
abulte: Probably! I'll see if I can refactor that later (if you didn't do it first ;-))

PR #33, opened 2023-07-06
Filter archived datasets

abulte:
Use another endpoint to fetch an organization's datasets which does not include archived datasets.

We need the organization technical id for this endpoint, so the fetch logic is a bit more convoluted.


src/services/api/resources/DatasetsAPI.js
```diff
@@ -2,4 +2,17 @@ import DatagouvfrAPI from "../DatagouvfrAPI"
 
 export default class DatasetsAPI extends DatagouvfrAPI {
   endpoint = "datasets"
+
+  /**
+   * Get datasets for an organization
+   *
+   * @param {str} org_id
```
streino, line 9: Specify that `org_id` must be the technical id here as well?

PR #38, opened 2023-07-12
Search improvements

abulte:
Adds pagination and organization facet to search page.

Needs https://github.com/ecolabdata/ecospheres-search-index/pull/2, which also implements a default sort.

I used a select for the organization facet, which is not optimal but there's no autocomplete component (for example) in DSFR.


streino:
Can the Org facet select be sorted?

abulte:
@streino It is! By dataset count.

streino:
Saw that after asking 😓 
I think it'd make more sense alphabetically. Freq sort makes it difficult to look for a specific org.

streino (approved):

PR #46, opened 2023-10-02
Bouquets: read views

abulte:
Adds a list and detail views for bouquets (aka topics).

Also improves some markdown / clamping rendering.

### DIFF
--- a/src/store/BouquetStore.js
+++ b/src/store/BouquetStore.js
@@ -18,7 +18,7 @@ export const useBouquetStore = defineStore("bouquet", {
       return bouquets.filter(bouquet => bouquet.extras.is_ecospheres || bouquet.tags.includes("ecospheres"))
     },
     /**
-     * Load Ecospheres related topics from API
+     * Load Ecospheres related bouquets from API
      *
      * @returns {Array<object>}
      */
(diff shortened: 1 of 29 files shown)


src/store/BouquetStore.js
```diff
@@ -0,0 +1,56 @@
...
+  }),
+  actions: {
+    /**
+     * Filter a list of bouquets related to Ecospheres
+     *
+     * @param {Array} bouquets
+     * @returns {Array}
+     */
+    filter (bouquets) {
+      return bouquets.filter(bouquet => bouquet.extras.is_ecospheres || bouquet.tags.includes("ecospheres"))
```
streino, line 18: We probably don't want to display the main "ecospheres" bouquet, especially when it'll contain thousands of datasets...

src/store/BouquetStore.js
```diff
@@ -0,0 +1,56 @@
...
+    },
+    /**
+     * Load Ecospheres related bouquets from API
+     *
+     * @returns {Array<object>}
+     */
+    async loadBouquets () {
```
streino, line 25: So French! But `loadBunches`... We may have to find another term 😅
(not a change request, just me wondering how i18n is going to work when we need it)

src/store/BouquetStore.js
```diff
@@ -0,0 +1,56 @@
...
+      if (this.data.length > 0) return this.data
+      let response = await topicsAPI._list()
```
streino, line 27: We'll need a way to pass a filter to data.gouv otherwise this won't scale.

streino (approved):

src/store/BouquetStore.js
abulte: True, I guess we can put the Ecospheres' bouquet id in the config file when it exists, and filter it out. Or add an extra `is_universe` or something.

src/store/BouquetStore.js
abulte: Yep :p Bouquet is legit though, I checked 🇬🇧

src/store/BouquetStore.js
streino: Will add it to our shared backlog with data.gouv

PR #47, opened 2023-10-03
Add bouquet creation and edition

abulte:
Add bouquet creation and edition.

Also refactors the API wrapper a bit to handle more than GET requests.

@streino sorry the commit history is a mess :-/ I should have rebased.


src/store/BouquetStore.js
```diff
@@ -47,10 +47,36 @@ export const useBouquetStore = defineStore("bouquet", {
...
+     *
+     * @param {string} bouquet_id
+     * @param {object} data
+     * @returns {object}
+     */
+    async update (bouquet_id, data) {
+      const res = await topicsAPI.update(bouquet_id, data)
+      this.data = [
+        ...this.data.filter(b => b.id !== bouquet_id),
+        res,
```
streino, line 77: This changes display order in the bouquet list view.
If it's simple to add alphabetical sort, let's do it, otherwise we'll track it as TODO.

src/views/bouquets/BouquetEditView.vue
```diff
@@ -0,0 +1,147 @@
...
+const search = async (query) => {
+  if (!query) return []
+  const results = await searchAPI._search(query, { hitsPerPage: 10 })
+  return results.hits.map(r => {
+    return { value: r.id, label: r.title }
+  }).filter(r => !datasets.value.map(d => d.id).includes(r.value))
+}
+
+const onSelect = async (value) => {
+  const dataset = await datasetStore.load(value)
```
streino, line 32: I get an error when selecting a dataset that doesn't exist on data.gouv (I'm guessing search index lag).
For instance, search "eau" and select "Donnée Sécheresse - Propluvia" => https://demo.data.gouv.fr/api/1/datasets/6470b39cfbad66b8c265ada3/ isn't found and breaks.

src/views/bouquets/BouquetEditView.vue
```diff
@@ -0,0 +1,147 @@
...
+    res = await bouquetStore.create({
+      ...data,
+      tags: ["ecospheres"],
+      extras: { is_ecospheres: true },
+    })
+  } else {
+    res = await bouquetStore.update(loadedBouquet.value.id, {
+      ...data,
+      tags: loadedBouquet.value.tags,
+    })
```
streino, line 58: `extras` is preserved, right?

src/services/api/DatagouvfrAPI.js
```diff
@@ -23,25 +37,40 @@ export default class DatagouvfrAPI {
...
+   * Base function for HTTP calls
    *
    * @param {string} url
-   * @returns {import("../../composables/fetch").ComposableFetchResult}
+   * @param {string} method
+   * @param {object} params
+   * @returns
    */
-  async makeRequestAndHandleResponse (url) {
+  async fetch (url, method = "get", params = {}) {
```
streino, line 47: I associate "fetch" more with GET than other verbs: "Fetch a POST ?"
How about a more generic "request"?

src/views/bouquets/BouquetEditView.vue
abulte: yes, if we don't touch it it should be preserved.

src/views/bouquets/BouquetEditView.vue
abulte: Right, let's say error handling has not been a priority on this POC 😅 I fixed it to not crash the whole app when it happens. BTW I can't reproduce this way, I've reindexed on demo and I don't have this dataset.

src/store/BouquetStore.js
abulte: Good eye 👁️ I did it in order to preserve the original order instead.

src/views/bouquets/BouquetEditView.vue
streino: Yeah, not crashing the whole app sounds like a good middle ground ;)
I suspected my exemple would be short lived, good thing it was easy to diagnose

streino (approved):

PR #53, opened 2023-10-09
User description when adding a dataset to a bouquet

abulte:
A user description can be filled when adding a dataset to a bouquet. It can be edited later. It's also displayed on the bouquet detail view for each dataset.

NB: storage is done "flat" in the `extras`, using a specially formatted key for each dataset. It could be done via a nested list attribute `ecospheres: []` but it made the code more complex so I kept with flat.

### DIFF
--- a/src/views/bouquets/BouquetDetailView.vue
+++ b/src/views/bouquets/BouquetDetailView.vue
@@ -47,6 +47,7 @@ onMounted(() => {
           :description="d.description"
           :img="d.organization.logo"
           :is-markdown="true"
+          :notice="bouquet.extras[`ecospheres:${d.id}:description`]"
         />
       </li>
     </ul>
--- a/src/views/bouquets/BouquetEditView.vue
+++ b/src/views/bouquets/BouquetEditView.vue
@@ -16,9 +16,31 @@ const isCreate = route.name === "bouquet_add"
 
 const form = ref({})
 const datasets = ref([])
-const selectedDataset = ref(null)
+const selectedDataset = ref({})
 const selector = ref(null)
 const loadedBouquet = ref({})
+const isModalOpen = ref(false)
+const isEditDesc = ref(false)
+
+const modalActions = [
+  {
+    label: "Valider",
+    onClick: (event) => {
+      event.preventDefault()
+      onSubmitModal()
+      isModalOpen.value = false
+    }
+  },
+  {
+    label: "Annuler",
+    secondary: true,
+    onClick: (event) => {
+      event.preventDefault()
+      isModalOpen.value = false
+      selectedDataset.value = {}
+    }
+  },
+]
 
 const search = async (query) => {
   if (!query) return []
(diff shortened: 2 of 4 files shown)


src/views/bouquets/BouquetEditView.vue
```diff
@@ -28,43 +50,73 @@ const search = async (query) => {
...
+  isEditDesc.value = true
+  selectedDataset.value.id = dataset.id
+  selectedDataset.value.title = dataset.title
+  selectedDataset.value.description = datasets.value.find(d => d.dataset.id === dataset.id).description
+}
+
+const extrasFromDatasets = () => {
+  const extras = {}
+  for (const dataset of datasets.value) {
+    if (dataset.description) {
```
streino, line 84: Fails to cover case where description exists and is removed during edit.

src/views/bouquets/BouquetDetailView.vue
```diff
@@ -47,6 +47,7 @@ onMounted(() => {
           :description="d.description"
           :img="d.organization.logo"
           :is-markdown="true"
+          :notice="bouquet.extras[`ecospheres:${d.id}:description`]"
```
streino, line 50: Might be worth having a helper to generate extra key to avoid inconsistencies.

BTW, I could see some topics still have the nested extras you played with before settling on flat keys. The "bouquetti" topic for instance:
```
{
	"extras": {
		"ecospheres": [
			{
				"datasetId": "5a5f4f6c88ee387da4d252a3",
				"descriptions": "with desc"
			}
		],
		"ecospheres:5a5f4f6c88ee387da4d252a3:description": "parceque"
	}
}
```

streino (commented):
_Not specific to this PR but not sure it's worth creating an issue for at this point of dev_
You can add the same dataset twice in a bouquet. Search lets you do it, but saving the bouquet then fails with PUT returning 400.

abulte:
> Not specific to this PR but not sure it's worth creating an issue for at this point of dev
You can add the same dataset twice in a bouquet. Search lets you do it, but saving the bouquet then fails with PUT returning 400.

I had a safeguard on that but it broke when I added the description :-/ Fixed in #56

src/views/bouquets/BouquetDetailView.vue
abulte: Agreed but refactored in #56 so I suggest we tackle this later on, merging would not be fun

abulte: And yeah, we'll need to cleanup some topics :p

src/views/bouquets/BouquetEditView.vue
abulte: True. I fixed in the simplest way possible, but now we're storing empty values for each and every dataset even if no description is set. Not a big problem at this stage I guess, and it avoids complexity when saving the extras while preserving the previous ones. This could be better if we have a list or a dict dedicated to Ecospheres in extras.

streino (approved):

PR #54, opened 2023-10-09
Add linter config

maukoquiroga:
#### Technical changes

- Add `eslint` and `prettify` configurations.

#### Non-technical changes

- Update `README.md`


abulte:
Good idea 👍 

Maybe use [standardjs](https://standardjs.com) instead to be consistent with current style?

streino (dismissed):

PR #102, opened 2023-10-17
feat(topics): allow non admin to create and edit

abulte:
Supersedes #95 

- Show bouquet creation button to everyone
- Show edit button to admin or owner of the bouquet

### DIFF
--- a/src/views/bouquets/BouquetDetailView.vue
+++ b/src/views/bouquets/BouquetDetailView.vue
@@ -23,6 +23,13 @@ const goToEdit = () => {
   router.push({name: "bouquet_edit", params: { bid: bouquet.slug }},)
 }
 
+const canEdit = computed(() => {
+  return userStore.isAdmin() || (
+    userStore.$state.isLoggedIn &&
+    bouquet.value.owner?.id === userStore.$state.data?.id
+  )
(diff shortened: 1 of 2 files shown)


streino (approved):
FYI @YeLnatSs @maukoquiroga

maukoquiroga:
> The build is randomly failing on Netlify (see below), I think we might have introduced a bug somewhere :-/ The build is actually fine, I reran it with clearing cache, but it's not updating the check status.
> 
> ```
> 6:07:40 PM: > ecospheres-front@0.0.0 build
> 6:07:40 PM: > vite build
> 6:07:40 PM: failed to load config from /opt/build/repo/vite.config.js
> 6:07:40 PM: error during build:
> 6:07:40 PM: Error: The package "@esbuild/linux-x64" could not be found, and is needed by esbuild.
> 6:07:40 PM: If you are installing esbuild with npm, make sure that you don"t specify the
> "--no-optional" or "--omit=optional" flags. The "optionalDependencies" feature
> 6:07:40 PM: of "package.json" is used by esbuild to install the correct binary executable
> 6:07:40 PM: for your current platform.
> ```

I got the same issue this morning.

PR #253, opened 2023-12-11
fix: :bug: edited router call for FormMF to avoid crash in prod

Jorek57:
Close https://github.com/ecolabdata/ecospheres-front/issues/252


geoffreyaldebert (approved):
damn ! it was that that bothered me so much last friday...

PR #317, opened 2024-01-08
config: meteodatagouvfr

geoffreyaldebert:
Config : rework meteo.data.gouv.fr to get back to main code with actual prod config

No impact for ecospheres


maudetes (approved):
Looks good

PR #433, opened 2024-04-19
Fix mf new datasets

nicolaskempf57:
This PR reverses the `!` condition and fixes the case for search with period but without department.


Pierlou (approved):
La review me fait dire qu'on est bons ! Merci 🙏

PR #603, opened 2024-11-26
Fix result number on search dataset

at-github:
Fix: https://github.com/ecolabdata/ecospheres/issues/449


DanFluture275 (approved):
testé PO OK

PR #619, opened 2024-12-10
chore: Add orgs to ecospheres demo universe

streino:
Fix ecolabdata/ecospheres#489
Fix ecolabdata/ecospheres#504

- (atmo-occitanie missing on demo)
- institut-national-de-lenvironnement-industriel-et-des-risques
- rnb

In case we don't switch to Grist on the next MEP.


abulte:
https://github.com/opendatateam/udata-front-kit/pull/613 mergé

PR #637, opened 2024-12-17
fix: bouquetCard

narduin:
conditional render of tags on bouquetCard

fix: https://github.com/ecolabdata/ecospheres/issues/515


abulte (approved):

DanFluture275 (approved):
testé PO OK

PR #647, opened 2025-01-15
fix: searchComponent

narduin:
Fix wrong dropdown position when dropdown opens up (drownup?).

Fix https://github.com/ecolabdata/ecospheres/issues/522


abulte (approved):

DanFluture275 (approved):
testé PO OK

PR #683, opened 2025-03-05
fix: dataset resources list condition

abulte:


src/components/datasets/ResourcesList.vue
```diff
@@ -93,7 +93,7 @@ onMounted(async () => {
 </script>
 
 <template>
-  <template v-if="resources.length">
+  <template v-if="dataset.resources.total">
```
streino, line 96: sneaky...

streino (approved):

PR #748, opened 2025-05-16
fix: edit dataset in topic w/o existing purpose

abulte:
Fix https://github.com/opendatateam/udata-front-kit/issues/727


streino (approved):
OK for me but let's have @johanricher confirm

abulte:
@johanricher you should be able to test on https://deploy-preview-748--logistique.sandbox.data.developpement-durable.gouv.fr/?

johanricher:
Login doesn't work on deployments (or just this one?) so I can't fully review the changes.

abulte:
oh, we should fix that.

abulte:
hum, it's on prod and wildcard are not allowed there for oauth clients urls... let's merge?

PR #783, opened 2025-07-08
ci: reduce automatic review app deployments and add manual deployment option

bolinocroustibat:
# Reduce automatic review app deployments and add manual deployment option

## Context

Several developers were frustrated by the excessive number of preview deployments triggered on every commit. The previous workflow deployed review apps to all configured sites (ecospheres, meteo-france, logistique) automatically on every push of a PR, even when developers were only working on one specific site. This created unnecessary resource consumption and deployment noise.

### 🔧 Workflow improvements

- ~~**Removed `synchronize` trigger**: No more automatic deployments on every commit~~
- **Changed trigger strategy**: Keep `synchronize` but only redeploy if review app already exists (manual-first approach)
- **Added manual deployment option**: New `workflow_dispatch` trigger with GUI options
- **Smart update logic**: Check if review app exists before automatic redeployment

### 📝 Documentation updates

- **Added comprehensive deployment section**: Clear explanation of both preview and production deployment processes
- **Organized by deployment type**: Preview deployments first (more frequent), then production deployments
- **Added visual indicators**: Emojis and clear structure for better readability
- **Referenced source files**: Links to workflow files instead of duplicating configuration

## Benefits

✅ **Reduced resource consumption**: No unnecessary deployments
✅ **Better developer experience**: Deploy only when needed
✅ **Flexible deployment**: Manual option for any PR to any site
✅ **Clear documentation**: Complete guide for all scenarios

Fix https://github.com/ecolabdata/ecospheres/issues/566

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -79,6 +79,7 @@ onMounted(() => {
 
 const logoService = config.website.service_logo
 const logoText = config.website.rf_title
+const serviceTitle = config.website.title
 const logoOperator = config.website.logo_operator?.src
 const logoOperatorHeight = config.website.logo_operator?.header?.height
 const logoOperatorWidth = config.website.logo_operator?.header?.width
@@ -159,6 +160,7 @@ watch(
     :desc-text="footerPhrase"
     :ecosystem-links="footerExternalLinks"
     :mandatory-links="footerMandatoryLinks"
+    :home-title="`Retour à l'accueil du site - ${serviceTitle}`"
   />
 </template>
 
--- a/src/components/SearchComponent.vue
+++ b/src/components/SearchComponent.vue
@@ -104,6 +104,7 @@ const clear = () => {
       type="search"
       :label-visible="labelVisible"
       :placeholder="!labelVisible ? searchLabel : undefined"
+      :title="!labelVisible ? searchLabel : undefined"
     >
       <template #before-input>
         <VIconCustom name="search-line" class="search-icon" />
@@ -111,6 +112,7 @@ const clear = () => {
     </DsfrInputGroup>
   </search>
 
+  <!-- FIXME: nested input needs a title attribute for a11y, but not supported by DsfrSearchBar -->
   <DsfrSearchBar
     v-else-if="!dropdown.length"
     v-model="query"
@@ -141,7 +143,8 @@ const clear = () => {
       :clear-on-search="true"
       :clear-on-blur="true"
       :close-on-select="true"
-      placeholder="Rechercher"
+      :placeholder="searchLabel"
+      :title="searchLabel"
       aria-owns=""
       :aria="{
         'aria-describedby': `${id}-description`,
--- a/src/components/forms/ErrorSummary.vue
+++ b/src/components/forms/ErrorSummary.vue
@@ -21,10 +21,11 @@ defineProps({
   <div
     v-show="formErrors.length"
     class="fr-my-4w fr-p-2w error-status"
-    role="group"
+    role="alert"
     aria-labelledby="error-summary-title"
+    tabindex="-1"
   >
-    <component :is="headingLevel" id="error-summary-title" tabindex="-1">
+    <component :is="headingLevel" id="error-summary-title">
       Il y a {{ formErrors.length }} erreur<span v-if="formErrors.length > 1"
         >s</span
       >
--- a/src/components/header/HeaderComponent.vue
+++ b/src/components/header/HeaderComponent.vue
@@ -14,7 +14,7 @@ type Props = {
 withDefaults(defineProps<DsfrHeaderProps & Props>(), {
   operatorImgAlt: '',
   operatorImgStyle: () => ({}),
-  searchLabel: 'Recherche',
+  searchLabel: 'Rechercher',
   quickLinks: () => [],
   showSearch: config.website.header_search.display,
   customSearch: false
@@ -58,6 +58,7 @@ const dropdown = config.website.header_search.dropdown ?? undefined
     service-description=""
     :quick-links
     :show-search="showSearch && !customSearch"
+    :home-label="`Retour à l'accueil du site - ${serviceTitle}`"
   >
     <!-- needed because of logo + badge -->
     <template #operator>
@@ -93,7 +94,7 @@ const dropdown = config.website.header_search.dropdown ?? undefined
         class="custom-search"
         :search-label="searchLabel"
         :dropdown="dropdown"
-        placeholder="Rechercher"
+        :placeholder="searchLabel"
         @do-search="closeModal"
       />
     </template>
--- a/src/views/topics/TopicFormView.vue
+++ b/src/views/topics/TopicFormView.vue
@@ -180,11 +180,11 @@ const onSubmit = async () => {
   await formFields.value.onSubmit()
   if (formErrors.value.length > 0) {
     setTimeout(() => {
+      errorSummary.value.$el.focus({ preventScroll: true })
       errorSummary.value.$el.scrollIntoView({
         behavior: 'smooth',
         block: 'start'
       })
-      errorSummary.value.$el.focus()
     }, 0)
   } else {
     save()


bolinocroustibat:
> Making sure I understand: All sites listed in the `inputs` are automatically deployed once on new PRs targeting main. Then they're not updated on new commits. If a dev wants to update, they have to manually do it.
> 
> Am I correct? If yes, then:
> 
> I agree the auto-deploys are noisy, but I'm worried we'll forget to manually update deployments and end up testing on stale versions. Especially if it's someone more on the product side doing the testing, and the dev forgot to update.
> 
> I'd be tempted to reverse the auto/manual logic: Nothing is automatically deployed when the PR is opened. The PR author has to manually deploy the site(s) they want. But once a site is manually deployed on a PR, it is automatically redeployed on new commits to ensure it's always up-to-date. Not sure how easy it is to implement, but it'd be safer that way. WDYT?

Yes you're correct. Actually, this PR is a proposal, but since several teams work with this repo and we don't all know about their needs, I'm very open to your feedbacks and proposal for different CI policies, as you just did.

Your proposal makes a lot of sense, I agree. I'm not sure how to implement that neither but I'll have a quick look now

.github/workflows/review-app.yml
```diff
@@ -91,7 +103,7 @@ jobs:
           ssh_private_key: ${{ secrets.REVIEW_APP_SSH_PRIVATE_KEY }}
 
       - name: Set site id as build arg
-        if: always()
+        if: always() && (github.event_name == 'workflow_dispatch' || env.REVIEW_APP_EXISTS == 'true')
```
streino, line 106: Don't need `always` here anymore

streino (dismissed):
Looks good. Thanks for the PR!

abulte:
FYI, I have a similar draft here https://github.com/opendatateam/udata-front-kit/pull/736

streino:
> FYI, I have a similar draft here #736

Je l'avais oublié celle là... Je sais pas sur quelle base il vaut mieux partir, mais ce serait pas mal d'adopter le `/deploy` dans tous les cas !

abulte (commented):
Thanks for having a go at this! I know the pain it is to work on gh workflows :-)

I'm not sure the `reopened` logic works. When a PR is reopened, it will have gone through the `closed` logic, this `REVIEW_APP_EXISTS` will be false. I think we need to go through the workflow dispatch step to recreate the app, and then `synchronized` will work as intended.

abulte:
> > FYI, I have a similar draft here #736
> 
> Je l'avais oublié celle là... Je sais pas sur quelle base il vaut mieux partir, mais ce serait pas mal d'adopter le `/deploy` dans tous les cas !

Si on part sur #783, on devrait pouvoir ajouter le trigger plus tard. Pas la peine de trop complexifier cette PR.

.github/workflows/review-app.yml
```diff
@@ -18,93 +36,133 @@ jobs:
...
+            echo "SITE=${{ matrix.site }}" >> $GITHUB_ENV
+            echo "PR_NUMBER=${{ github.event.pull_request.number }}" >> $GITHUB_ENV
+            echo "EVENT_ACTION=${{ github.event.action }}" >> $GITHUB_ENV
+            # Set defaults (no deployment, no creation)
+            echo "SHOULD_DEPLOY=false" >> $GITHUB_ENV
+            echo "SHOULD_CREATE=false" >> $GITHUB_ENV
+            
+            # Override defaults based on event type
+            if [ "${{ github.event.action }}" = "reopened" ] || [ "${{ github.event.action }}" = "synchronize" ]; then
+              # For reopened/synchronized PRs, only redeploy if review app already exists (manual-first approach)
```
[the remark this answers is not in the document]
abulte: It can if the deletion failed (happens sometimes under heavy load). Not nominal, obviously.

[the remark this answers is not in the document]
bolinocroustibat: I'm also q bit confused on the behaviour of GitHub Actions for a reopened PR. Should we just remove any logic related to "reopened", and assume the review app will be created manually though the CI UI? That would make sense and be simpler indeed.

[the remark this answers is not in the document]
streino: I'd drop "reopened" on this PR to keep things simple. If it ever becomes too much of a chore to manually re-deploy re-opened PRs, we can revisit. It'll require flagging the "previously deployed" state somewhere. Since we're deploying on a single host we could probably store it there as a simple touch file, but I'm not convinced the complexity is worth it.

.github/workflows/review-app.yml
```diff
@@ -18,93 +36,138 @@ jobs:
...
-          # - simplifions
+          - defis
+          - hackathon
+          - simplifions
     permissions:
       deployments: write
-    # only run if the PR is from the same repo
-    if: github.event.pull_request.head.repo.full_name == github.repository && github.event_name == 'pull_request'
+    # only run if PR is from same repo OR if manually triggered
+    if: (github.event_name == 'pull_request' && github.event.pull_request.head.repo.full_name == github.repository) || github.event_name == 'workflow_dispatch'
```
streino, line 45: So, for a PR coming from a different repository, the user will have to manually trigger deploy updates. This may be fine, just want to check it's what we expect.

README.md
```diff
@@ -66,9 +66,33 @@ npm run format
 
 ## Déploiement
 
-### Workflow GitHub pour le déploiement
+### 🔍 Déploiement en preview
 
-Le déploiement des verticales thématiques s'effectue via un workflow GitHub qui se déclenche automatiquement à partir du message de commit. Le format du message de commit doit être :
+Une **review app** est un environnement de prévisualisation temporaire qui permet de tester les changements d'une Pull Request dans un environnement similaire à la production. Un workflow CI/CD dédié ([`.github/workflows/review-app.yml`](.github/workflows/review-app.yml)) gère automatiquement la création, mise à jour et suppression de ces environnements.
+
+Les **review apps** ne sont **plus créées automatiquement** lors de l'ouverture d'une Pull Request. L'auteur de la PR doit **déployer manuellement** les PR qu'il souhaite tester, via l'interface de GitHub Actions.
```
streino, line 73:
```suggestion
Les **review apps** ne sont **pas créées automatiquement** lors de l'ouverture d'une Pull Request. L'auteur de la PR doit **déployer manuellement** les PR qu'il souhaite tester, via l'interface de GitHub Actions.
```

README.md
```diff
@@ -66,9 +66,33 @@ npm run format
...
+
+> **💡 Info** : Une fois qu'une review app est créée pour une PR, elle sera **automatiquement mise à jour** à chaque nouveau commit sur la PR. Si une PR est fermée puis rouverte, une nouvelle review app sera automatiquement créée.
```
streino, line 75: Outdated

streino (commented):
Looks good! Some misc comments and we should be good to go!

.github/workflows/review-app.yml
bolinocroustibat: This is was you were suggesting as the manual approach first, right? This makes sense to me but I'm not familiar with the use of this repo by its contributor, so I'll leave it to reviewers to confirm!

streino: I do confirm "manual first" is OK, but here I'm specifically talking about a PR coming from a different repo.

If PR comes from the same repo, then "manual first" works as expected : 
- First, someone manually triggers the deploy : `github.event_name == 'workflow_dispatch'` => OK.
- Then, each PR update automatically triggers a redeploy : `github.event_name == 'pull_request' && github.event.pull_request.head.repo.full_name == github.repository` => OK.

However, if PR comes from a different repo : 
- First, someone manually triggers the deploy => OK.
- Then, each PR update automatically triggers a redeploy => **KO**, because `github.event.pull_request.head.repo.full_name != github.repository`.

I'm thinking we might want sync to always automatically redeploy. Once a deployment has been manually triggered, it seems better (least surprise) to keep those deployments up to date?

PR #790, opened 2025-07-17
fix(rgaa): add topic title

abulte:
Fix https://github.com/ecolabdata/ecospheres/issues/666

Capture d’écran 2025-07-17 à 11 56 41

### DIFF
--- a/src/views/topics/TopicFormView.vue
+++ b/src/views/topics/TopicFormView.vue
@@ -223,6 +223,8 @@ onMounted(() => {
         setMetaTitle()
       })
       .finally(() => loader.hide())
+  } else {
+    setMetaTitle()
   }
 })
 </script>


src/views/topics/TopicFormView.vue
```diff
@@ -223,6 +223,8 @@ onMounted(() => {
         setMetaTitle()
       })
       .finally(() => loader.hide())
+  } else {
```
streino, line 226: Since you're calling it in both paths, can you just move it after the top-level `if`?
```
  if (!props.isCreate || routeQuery.clone != null) {
    ...
  }
  setMetaTitle()
```

abulte: Nope, cause the first branch is async and we want to apply at the end of the flow.

streino (approved):

PR #802, opened 2025-07-22
feat(ci): add comment based deploy

abulte:
Permet de lancer le workflow `review-app` depuis un commentaire du type `/deploy {site}`.

J'ai choisi de faire au plus simple dans un premier temps : ne supporte qu'un seul site et délégue la validation à l'autre workflow.

Il faut malheureusement merger sur main pour pouvoir tester :-/


.github/workflows/comment-deploy.yml
```diff
@@ -0,0 +1,79 @@
...
+      branch: ${{ steps.get_pr.outputs.branch }}
+    steps:
+      - name: Parse comment
+        id: parse
+        run: |
+          COMMENT="${{ github.event.comment.body }}"
+          echo "Comment: $COMMENT"
+
+          # Extract /deploy command and site
+          if [[ "$COMMENT" =~ /deploy[[:space:]]+([a-zA-Z-]+) ]]; then
```
streino, line 24:
```suggestion
          if [[ "$COMMENT" =~ /deploy[[:space:]]+([a-zA-Z-_]+) ]]; then
```

.github/workflows/comment-deploy.yml
```diff
@@ -0,0 +1,79 @@
...
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          script: |
+            const pr = await github.rest.pulls.get({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              pull_number: ${{ steps.parse.outputs.pr_number }}
+            });
+
+            console.log('PR branch:', pr.data.head.ref);
+            core.setOutput('branch', pr.data.head.ref);
```
streino, line 49: Equivalent to `echo ... >> $GITHUB_OUTPUT` above? If yes, we might want stick to a single way.
Same for `console.log` vs simple `echo`.

streino (dismissed):

streino (dismissed):

bolinocroustibat (dismissed):
I didn't think about the PR comments, that looks quite practical, thanks!

Maybe we could use also PR comments instead of commits names/tag names as well to deploy programmatically to preprod/prod in [this PR](https://github.com/opendatateam/udata-front-kit/pull/803), so that we have consistent and easy-to-maintain deploy means across different workflows? What do you think?

bolinocroustibat (approved):
Seinfeld Pez

PR #803, opened 2025-07-22
ci: refactor CI for deploying on preprod and prod

bolinocroustibat:
## 🔄 Refactor deployment workflow to manual trigger

**Context:**
- Developers complained about CI noise from unnecessary pipeline triggers
- Non-technical users struggled with commit message format as deployment barrier
- Manual trigger provides better accessibility and readability

**Changes:**
- Replace commit message-based deployment with manual GitHub Actions interface
- Add GUI inputs for site, environment (preprod/prod), and version type (major/minor/patch)
- Remove automatic trigger on push, now requires manual workflow dispatch
- Update README with clear deployment instructions

The workflow still calculates the next version and tag automatically, and triggers GitLab deployment via the scaffold script.


.github/workflows/create-deploy-release.yml
```diff
@@ -1,71 +1,107 @@
...
       GITLAB_API_TOKEN: ${{ secrets.GITLAB_API_TOKEN }}
       DEPLOY_PRIVATE_KEY: ${{ secrets.DEPLOY_PRIVATE_KEY }}
+      
+      # workflow dispatch inputs
+      CONFIG_NAME: ${{ github.event.inputs.site }}
+      ENV: ${{ github.event.inputs.environment }}
+      VERSION_PART: ${{ github.event.inputs.version_type }}
     steps:
-      - name: configure environment
+      - name: Debug event and inputs
```
jordanguedj, line 60: This is not needed I think, it's already displayed on the GA UI:

.github/workflows/create-deploy-release.yml
```diff
@@ -1,71 +1,107 @@
...
           git config --global user.name "datagouv"
 
-      - name: clone repositories
+      - name: Clone this repo and the scaffold script repository
         shell: bash
         run: |
           git clone --quiet --depth 1 -b ${{ env.REPO_CURRENT_BRANCH }} ${{ env.REPO_SSH_URL }}
           git clone --quiet --depth 1 $SCAFFOLD_REPO_SSH_URL ${{ env.SCAFFOLD_DIR }}
 
-      - name: check commit message
```
jordanguedj, line 49: We should find a way to keep the possibility to trigger deployment using CLI.

.github/workflows/create-deploy-release.yml
```diff
@@ -1,71 +1,107 @@
-name: Create and deploy a new release
-run-name: create and deploy ${{ github.sha }}
+name: Deployment on preprod or prod with version bump
```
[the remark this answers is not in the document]
bolinocroustibat: This is the name of the workflow, I'm afraid unfortunately we cannot use the env vars here

JeSuisUnCaillou (approved):
JAPROUV

.github/workflows/create-deploy-release.yml
```diff
@@ -1,71 +1,107 @@
...
+
+      - name: Setup SSH keys and Git configuration
         shell: bash
         run: |
-          # configure environment variables
-          echo "REPO_HTTPS_URL=${{ github.server_url }}/${{ github.repository }}.git" >> $GITHUB_ENV
+          # Set Git repository URLs and branch info
           echo "REPO_SSH_URL=git@github.com:${{ github.repository }}.git" >> $GITHUB_ENV
           echo "REPO_CURRENT_BRANCH=${GITHUB_HEAD_REF:-${GITHUB_REF#refs/heads/}}" >> $GITHUB_ENV
-          echo "SCAFFOLD_DIR=scaffold" >> $GITHUB_ENV
```
jordanguedj, line 30: To check: it's possible that this variable should be included in `$GITHUB_ENV`.

.github/workflows/create-deploy-release.yml
```diff
@@ -1,71 +1,107 @@
...
+      SCAFFOLD_DIR: "scaffold"
       ENVS: ${{ vars.ENVS }}
       APPS: ${{ vars.APPS }}
 
       # secrets to be defined
       GITLAB_API_TOKEN: ${{ secrets.GITLAB_API_TOKEN }}
       DEPLOY_PRIVATE_KEY: ${{ secrets.DEPLOY_PRIVATE_KEY }}
+      
+      # workflow dispatch inputs
+      CONFIG_NAME: ${{ github.event.inputs.site }}
```
jordanguedj, line 56: Same here: we should check that those variables does not need to be included in `$GITHUB_ENV`.

jordanguedj (changes requested):
Thanks for this! It's a good improvement to allow non-tech users to deploy easily.
However, it's removing the possibility to deploy from CLI.
Also, I think it deserved to be tested properly before merge.

.github/workflows/create-deploy-release.yml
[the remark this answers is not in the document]
jordanguedj: Yes, then I think the name should be generic and should not specify environment names as it could differs between app, and we could add additional envs later.

.github/workflows/create-deploy-release.yml
abulte: Tags would be great :-)

streino:
> A bit more context sorry:
> 
> This is a PR done after a long discussion with @jordanguedj and @JeSuisUnCaillou , also taking into account the non-tech users feedbacks collected by @JeSuisUnCaillou.
> 
> A first remark on this PR by @jordanguedj is that we cannot deploy to preprod/prod programatically anymore, with git CLI, for dev users. We could actually have both. If that's a common need among devs, I suggest we can do that in a later PR?

Cool!

When we initially discussed this with @jordanguedj I also suggested using Github Releases as triggers for deployments. IMO Releases are more user-friendly than commit-based triggers (would they be user-friendly enough?), and they can be created both through the UI and the Github CLI [1]. Also, and I think this is important, they create a lasting trace of deployed versions, when Action-based deploys are harder to track.

We could use an Action to create a release on deploy (Deploy => Release), but since releases sometime require manual intervention (eg editing the release notes) I think the other way around (Release => Deploy) would work better.

[1]: We can't directly create a release from the standard git CLI, but we could automate the creation of Releases based on pushed tags. So we'd have Tag => Release + Deploy. This does raise the question of manual intervention on releases (cf previous paragraph).

Maybe an intermediate step is to go with the proposed workflow but only allowing the deployment of tags and/or releases (at least in prod)?

.github/workflows/create-deploy-release.yml
streino: +1 for (preferably named-) tags. See also https://github.com/opendatateam/udata-front-kit/pull/803#issuecomment-3103521879 on how we might want to articulate it with releases ?

bolinocroustibat: Great. Actually might make sense to include that in this PR, I will suggest something in a commit in same PR

streino: I'd rather we try to get things right in this PR to avoid changing the deployment process multiple times. But if people doing the actual deployments are OK with multiple changes, why not.

.github/workflows/create-deploy-release.yml
```diff
@@ -1,71 +1,99 @@
...
+          - defis
+          - hackathon
+          - simplifions
+      environment:
+        description: 'Environment to deploy to'
+        required: true
+        default: 'preprod'
+        type: choice
+        options:
+          - preprod
```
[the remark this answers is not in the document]
bolinocroustibat: Oh the env names not the same between sites? I would prefer standardizing them, as otherwise it's quite error-prone in the GUI non non-dev users if we have both "preprod" and "demo" listed in the choices.

streino:
Also, the combination of opening deployment to non-technical users and easier form-based selection might increase the risk of deploying to the wrong target. 
Ideally we'd want to manage per-vertical/env deploy rights but I don't know how far we could go with Github? Another option would be to add some sort of guard to prevent basic mistakes (could be having to type de vertical instead of selecting it, or having a confirmation field of some sort). If we go the tag/release way, having to type the name of the tag/release when creating it might be enough.

.github/workflows/create-deploy-release.yml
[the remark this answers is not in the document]
jordanguedj: Sure!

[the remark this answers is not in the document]
bolinocroustibat: Fine with me, fixed. Let's try for the CI not to be overcomplicated by not adding a condition in the CI for now.

.github/workflows/create-deploy-release.yml
```diff
@@ -65,10 +64,43 @@ jobs:
...
+            echo "Tag: ${{ github.ref_name }}"
+          fi
+
+      - name: Parse tag and set deployment variables
+        if: github.event_name == 'push' && startsWith(github.ref, 'refs/tags/')
+        run: |
+          TAG_NAME="${{ github.ref_name }}"
+          echo "Parsing tag: $TAG_NAME"
+          
+          # Parse tag format: {site}-{environment}-{version}
```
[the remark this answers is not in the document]
streino: BTW, the per-vertical config repos would solve [per-vertical deployment rights](https://github.com/opendatateam/udata-front-kit/pull/803#issuecomment-3103604607). So maybe we should seriously look into making that monorepo split happen?

bolinocroustibat:
> Also, the combination of opening deployment to non-technical users + easier form-based selection might increase the risk of deploying to the wrong target.
> 
> Ideally we'd want to manage per-vertical/env deploy rights but I don't know how far we could go with Github? Another option would be to add some sort of guard to prevent basic mistakes (could be having to type de vertical instead of selecting it, or having a confirmation field of some sort). If we go the tag/release way, having to type the name of the tag/release when creating it might be enough.
> 
> In general, I'd be worried about having non-technical users being able to deploy whatever commit they like, or even just HEAD. For "most recent deployment" we could let the user only deploy the HEAD of per-env dedicated branches, and retain some technical control through PR reviews of what gets merged in those branches. But it might be necessary to support roll-backs, and so having tags/releases seem again like the better way.

Yes those are good points, I briefly was concerned about that as well, although I generally tend to trust users, tech or not tech, if it's clearly explained.
My (too relaxed?) approach on this would be to test-drive this basic workflow as it is, collect feedback about it from users on the go, see if any danger or wrongdoing arise, and strengthen security later if needed - but I don't have the visibility on the number and profile of non-tech people using the repo now and in the future, so I very much might be proven wrong.

.github/workflows/create-deploy-release.yml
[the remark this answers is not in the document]
jordanguedj: We still don't want to let the verticals choose their versioning scheme, at least for now, as it implies allowing this and changing this for all the verticals. We want to ensure consistency between verticals versioning (we also want that git tags and docker image versions match and we want to keep that consistency between all the apps deployed on our infra). Also, this CI has been design to automate the git tag creation in order to: ensure that consistency, but also to avoid tag conflicts.

However I can understand the need, maybe we could discuss this (monorepo split as you suggested?).

.github/workflows/create-deploy-release.yml
jordanguedj: Thanks for this. However as it is, it will conflict will the next steps of the workflow, as tag names are [deducted](https://github.com/opendatateam/udata-front-kit/blob/main/.github/workflows/create-deploy-release.yml#L76) from `VERSION_PART` variable. Initially, this deployment workflow has been designed to automate the git tag creation in order to: ensure consistency between verticals versioning, but also to avoid tag conflicts.

.github/workflows/create-deploy-release.yml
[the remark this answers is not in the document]
streino: Makes sense, and yes let's put that on the list of things to consider when splitting the repo.

streino:
> [@bolinocroustibat] Yes those are good points, I briefly was concerned about that as well, although I generally tend to trust users, tech or not tech, if it's clearly explained. My (too relaxed?) approach on this would be to test-drive this basic workflow as it is, collect feedback about it from users on the go, see if any danger or wrongdoing arise, and strengthen security later if needed - but I don't have the visibility on the number and profile of non-tech people using the repo now and in the future, so I very much might be proven wrong.

To be clear, I'm not preemptively blaming non-tech users, but supposedly the tech people doing the deploys have some background and have learned to be careful with such things, if only because they were the one having to clean up the mess a few too many times. Explaining goes a long way for sure, but if we make it easy to deploy with just a few drop boxes, I'm not sure how many will read (and remember) the fine points in the doc.

I'm all for making things easier/simpler/more open, but I also like defensive systems. Yes it makes building the system more complex, but if you're aiming for user-friendliness, it shouldn't fall on the user to mind their steps.

> [@JeSuisUnCaillou] If I choose to deploy to ecologie by mistake (which is as likely for a tech than a non tech to be honest), wouldn't I just deploy ecologie's last version to the ecologie production ? It won't override any environnment variable, so I don't see how I'm gonna be able to deploy something wrong to another vertical.

First a deployment can always cause two things:
1. Downtime.
1. Potential deployment issues leaving the site broken.

Given we're dealing with a static site this isn't too concerning, but it's there.

Second, depends on what we can redeploy:
1. If we can only redeploy the latest commit, then:
    1. If site was already deployed on this commit, you end up redeploying the same. This is the "best case" for a non-expected redeploy: No impact on the version deployed, but you still get the potential downtime issues.
    1. If site was lagging behind the latest commit (for whatever reason, like scheduling a deployment at some later time), then you end up unexpectedly releasing the new version.
1. If we can redeploy older commits/tags (either by directly picking a commit/tag, or by re-running an old deployment action), then you end up unexpectedly releasing an old version.

---

Having said all that, my take would be to focus on moving away from the current monorepo rather than build too many defensive features now.

.github/workflows/create-deploy-release.yml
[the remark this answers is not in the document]
streino: Next steps here: https://github.com/opendatateam/udata-front-kit/issues/809

.github/workflows/deploy-release.yml
```diff
@@ -0,0 +1,156 @@
...
+
+      - name: Parse tag and set deployment variables
+        if: github.event_name == 'push' && startsWith(github.ref, 'refs/tags/')
+        run: |
+          TAG_NAME="${{ github.ref_name }}"
+          echo "Parsing tag: $TAG_NAME"
+          
+          # Parse tag format: {site}-{environment}-{version_type}
+          # Example: ecospheres-prod-patch
+          if [[ $TAG_NAME =~ ^([a-zA-Z0-9-]+)-(prod|preprod|demo)-(major|minor|patch)$ ]]; then
```
jordanguedj, line 85: Allowing tag name format `ecospheres-prod-<major, minor, patch>` will break consistency between git tags and Docker image version, which is what we would like to avoid.

.github/workflows/create-deploy-release.yml
jordanguedj: Thank you for your dedication to improve this deployment workflow.
However, allowing users to push tags with another format than `<app>-<env>-X.X.X` breaks the consistency between git tags and Docker image version, and we would like to avoid that (at least for this repository).

IMO it would be preferable to move forward step by step: make deployment from inputs is already a good improvement. This workflow could definitely be improved, and @streino suggestion of a monorepo split seems a good way to add more specific ways to deploy, and answer specific needs.

bolinocroustibat: OK, let's forget about git tags for now then, I'll put the git commit naming logic back instead along with GitHub GUI if that's OK for everyone

PR #812, opened 2025-07-24
fix: add workflow file for comment deploy

abulte:
I forgot to `git add` the meat of the work in #802 (because of file renaming) :-/


streino (approved):
:sweat_smile:

PR #814, opened 2025-07-24
fix: force page remount when switching detail pages

abulte:
Injecting `pageKey` in props forces the detail page to remount when switching page type.


JeSuisUnCaillou (approved):
Super ça marche, merci :pray:

PR #817, opened 2025-07-28
add config to display or hide datasets and reuses

JeSuisUnCaillou:


JeSuisUnCaillou:
Encore une PR que j'ajoute dans la chaîne de PRs  https://github.com/opendatateam/udata-front-kit/pull/810 et https://github.com/opendatateam/udata-front-kit/pull/815

configs/defis/config.yaml
```diff
@@ -324,6 +328,10 @@ pages:
     discussions:
       display: false
       create: false
+    datasets:
```
streino, line 331: I get where this comes from, but having "datasets" in a "datasets" section... We could name it "resources", "files" or something else, but it'd break the mapping with the API.

Let's keep it the way it is in this PR, but let's consider grouping all tab-related config later. @abulte WDYT?

configs/simplifions/config.yaml
```diff
@@ -232,6 +236,35 @@ pages:
             name: ⭐️⭐️ Dites-le nous une fois | L'usager n'a plus à fournir de justificatifs
           - id: acces-facile
             name: ⭐️ Accès facile | L'usager ou l'agent trouve l'information
+
```
streino, line 239: Out of PR scope

src/views/topics/TopicDetailView.vue
```diff
@@ -85,11 +85,25 @@ const breadcrumbLinks = computed(() => {
...
 })
 
-const tabTitles = [
-  { title: 'Données', tabId: 'tab-0', panelId: 'tab-content-0' },
-  { title: 'Discussions', tabId: 'tab-1', panelId: 'tab-content-1' },
-  { title: 'Réutilisations', tabId: 'tab-2', panelId: 'tab-content-2' }
-]
+const tabTitles: { title: string; tabId: string; panelId: string }[] = []
+if (pageConf.datasets.display === true) {
+  tabTitles.push({ title: 'Données', tabId: 'tab-0', panelId: 'tab-content-0' })
```
streino, line 90: The int indices in `tabId` and `panelId` don't make as much sense now. And could even be confusing since we could have "tab-2" at index 0.

Could we rename them to the actual contents? I.e "tab-0" -> "tab-datasets" etc.

src/views/topics/TopicDetailView.vue
```diff
@@ -417,13 +431,19 @@ watch(
...
       :tab-titles="tabTitles"
       :tab-list-name="`Groupes d'attributs du ${pageConf.labels.singular}`"
     >
       <!-- Jeux de données -->
-      <DsfrTabContent panel-id="tab-content-0" tab-id="tab-0" class="fr-px-2w">
+      <DsfrTabContent
+        panel-id="tab-content-0"
+        tab-id="tab-0"
+        class="fr-px-2w"
+        v-if="pageConf.datasets.display === true"
```
streino, line 445: Standardize on either `showXXX` or `pageConf.XXX.display` on all tabs.

src/views/topics/TopicDetailView.vue
JeSuisUnCaillou: Sure

JeSuisUnCaillou:
@streino Seems good for the rebase.

src/views/topics/TopicDetailView.vue
```diff
@@ -85,11 +87,29 @@ const breadcrumbLinks = computed(() => {
   return breadcrumbs
 })
 
-const tabTitles = [
-  { title: 'Données', tabId: 'tab-0', panelId: 'tab-content-0' },
-  { title: 'Discussions', tabId: 'tab-1', panelId: 'tab-content-1' },
-  { title: 'Réutilisations', tabId: 'tab-2', panelId: 'tab-content-2' }
-]
+const tabTitles: { title: string; tabId: string; panelId: string }[] = []
+if (pageConf.resources_tabs.datasets.display === true) {
```
streino, line 91:
```suggestion
if (showDatasets === true) {
```
Maybe even move the `=== true` in `showXXX`, so the `if` here the `v-if` below are guaranteed to work the same way?

configs/simplifions/config.yaml
JeSuisUnCaillou: Oh, right ! Sorry. I didn't go see the file as you were just pointing to one line.

src/views/topics/TopicDetailView.vue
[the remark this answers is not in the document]
JeSuisUnCaillou: I'll fix them.

But if that is mandatory to fix, it should be red here in the CI : 
image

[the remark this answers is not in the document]
streino: True. It was recently added in the diff view, so we haven't decided yet. CC @abulte

streino (approved):

PR #819, opened 2025-07-28
Display solutions editeurs + some fixes

JeSuisUnCaillou:


configs/simplifions/config.yaml
```diff
@@ -38,7 +38,7 @@ website:
     sections:
     meta_description:
   header_button:
-    display: true
+    display: false # TODO : créer un formulaire de feedback sur tally.so
     label: 'Donnez votre avis'
     link: https://tally.so/r/wLWrzp
```
streino, line 43: Maybe delete that one to make sure it's not incorrectly if `display` is reverted to true?

streino (approved):

configs/simplifions/config.yaml
JeSuisUnCaillou: can do

streino (approved):

PR #822, opened 2025-07-29
Simplifions minor fixes

JeSuisUnCaillou:


streino (dismissed):

streino:
> @streino Sorry I had to rebase, I need another validation now ^^'
> 
> I'm not sure how to do it differently, in order to not invalidate your review

Not sure either. If anyone knows of a setting that'll keep reviews on noop rebases, I can toggle it, but I think it's either always or never drop.

JeSuisUnCaillou:
> Not sure either. If anyone knows of a setting that'll keep reviews on noop rebases, I can toggle it, but I think it's either always or never drop.

There is one i'm sure, cause in my other repos, new commits don't invalidate a review. Let me look for it.

JeSuisUnCaillou:
Ah, but it will not check if the new commits are "just" a rebase with no modifications.

abulte:
@JeSuisUnCaillou can you add the doc for the `editable` conf up there https://github.com/opendatateam/udata-front-kit/blob/167244b96b42de03a04c7ee76611eb993f0037b8/configs/ecospheres/config.yaml#L195? 🙏

PR #846, opened 2025-08-06
chore: merge config at build time

abulte:


src/config.ts
```diff
@@ -1,31 +1,4 @@
...
-} else {
-  console.log(
-    `No specific config found for ${mode}, using default config of ${site_id}`
-  )
-}
-
-export default finalConfig
+// TODO: we should type the entire config here
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+declare const __APP_CONFIG__: any
```
streino, line 3: (iff this PR gets adopted)

Maybe add a comment about where this is set?

And nitpick: Do we need that many underscores or does `APP_CONFIG` work as well?

abulte: AFAIK it would work, it's just kind of a convention for internal stuff.

JeSuisUnCaillou (commented):
I couldn't figure out how to expose this merged json on init, thanks !

I'm pointing out one thing though (and this is me shooting myself in the foot) : `npm run dev` is not reacting to changes to the config anymore.

It's a small pain, but I'm starting to think it's not worth it. I'd rather fall back on my first approach, which is to use environment variables for the few values that depend on the environment.

JeSuisUnCaillou:
@jordanguedj told me you both are living around Paris : We could see each other one day at the office, and talk about all this in person, I feel like it would be much easier to get an actual understanding on each other's stakes.

PR #848, opened 2025-08-06
chore: Don't put too many backup owners, they get spammed

streino:


abulte (approved):
Maybe just remove from default, but nitpick.

PR #855, opened 2025-08-19
Fix dataset modal without editorialization

nicolaskempf57:
The validation is checking for fields hidden due to the `datasetEditorialization` prop


nicolaskempf57:
> A question : Why aren't all other values conditioned by the `if (props.datasetEditorialization)` ? (group, availability and availabilityUrl)

The group field is always shown, with or without `datasetEditorialization`.
I moved the availability ones inside the condition too to match the component logic in `DatasetPropertiesFields`.

JeSuisUnCaillou (approved):

abulte (approved):

PR #865, opened 2025-09-08
Use ViteRestart to detect config changes

streino:
Solution for https://github.com/opendatateam/udata-front-kit/pull/846#issuecomment-3249724862

Turns out we were not even restarting on base config...

### DIFF
--- a/vite.config.mts
+++ b/vite.config.mts
@@ -8,6 +8,7 @@ import ViteYaml from 'unplugin-yaml/vite'
 import { defineConfig, loadEnv } from 'vite'
 import dynamicImport from 'vite-plugin-dynamic-import'
 import { createHtmlPlugin } from 'vite-plugin-html'
+import ViteRestart from 'vite-plugin-restart'
 import vueDevTools from 'vite-plugin-vue-devtools'
 
 import {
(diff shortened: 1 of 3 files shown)


vite.config.mts
```diff
@@ -30,23 +31,28 @@ interface Config {
...
   const configDir = `./configs/${env.VITE_SITE_ID}`
+  const configFiles = []
 
   // Load base config
-  const configFileUrl = new URL(`${configDir}/config.yaml`, import.meta.url)
-  let config = load(readFileSync(configFileUrl, 'utf-8')) as Config
+  const mainConfigFile = `${configDir}/config.yaml`
+  configFiles.push(mainConfigFile)
+  let config = load(
+    readFileSync(new URL(mainConfigFile, import.meta.url), 'utf-8')
```
streino, line 40: Do we need URL here? Why not simply `mainConfigFile`?

JeSuisUnCaillou (approved):
Seems legit to me :+1: 

We can merge this into https://github.com/opendatateam/udata-front-kit/pull/846#issuecomment-3249724862 then into https://github.com/opendatateam/udata-front-kit/pull/828, and I'll work on the rebase then.

vite.config.mts
abulte: It's used to reliably resolve the file path to an absolute path, across platforms and in a ES module context. It might not be strictly required but I'd keep it for safety sake.

PR #878, opened 2025-09-16
simplifions v2

JeSuisUnCaillou:
Now simplifions will get the minimum information needed from the topics to display search results.
And to display a topic's page, we will now fetch the rest of the needed informations directly from Grist.


JeSuisUnCaillou:
@abulte Sorry for the sheer amount of modifications to review 😅

I have changed the behavior of simplifions to have :
- The topics now have almost no extras : only a grist ID + a few columns to feed the search result cards
- The detail views now request directly the Grist to display its data

This allows us to see live the modifications we make on the grist, on the detail views.

abulte (approved):

PR #893, opened 2025-10-01
feat(topic): activities list

abulte:
Fix https://github.com/ecolabdata/ecospheres/issues/601


abulte:
@streino @Thesauruv this is still a draft (I did not do as many tests as I need to, and the code needs a Birds Eye view), but the core features are demoable. Since we're on a deadline, I'll take feedback asap.

I might have gone overboard with the deeplinking system. Those are usually brittle, and this one is no exception. Please be gentle :-) I'll remove some parts of it if they're not stable enough and deemed not necessary.

Thesauruv (commented):
@abulte Thank you very much for this first implementation! In my opinion it is already functional. I see three improvements (by order of priority) : 
- I would keep the "Activités" tab visible in all cases, that is, whether the user is connected or not, and wheter they are a member of the organisation or not. The reason for this is that changes are valuable information for Bouquet users which are not direct contributors but still want to stay informed about the latest updates. If this point can be handled before the production release, that would be great.
- better highlight clickable links, for instance by adding an arrow at the end ? 

image

- Replacing factor IDs with their titles (optional for this MEP IMO)

Thesauruv:
Just to know if this is an expected in this first version.

When I change a group name, the extra of the first factor of this group is notified as changed.

image

image

cypress/e2e/custom/ecospheres/topics/elements/deeplink.cy.ts
```diff
@@ -0,0 +1,186 @@
...
+  let testTopic: Topic
+  let testFactors: Factor[]
+
+  beforeEach(() => {
+    ;({ testTopic, testFactors } = setupTopicWithExistingFactors())
+  })
+
+  describe('Navigating to a specific factor via URL hash', () => {
+    it('should navigate to a factor in a group when hash is present in URL', () => {
+      const targetFactor = testFactors[0]
```
streino: Could some of the assertions below be true even if the feature doesn't work just because it's the first factor, and e.g. has a good chance to already be in the viewport, or first group is already expanded by default (if we were to change expansion rules)?

cypress/e2e/custom/ecospheres/topics/elements/deeplink.cy.ts
```diff
@@ -0,0 +1,186 @@
...
+      cy.wait('@getElementsDataset')
+
+      cy.get(`#factor-${firstFactor.id}`).isInViewport({ wait: VIEWPORT_WAIT })
+      cy.contains(firstFactor.title).should('be.visible')
+
+      // Now navigate to second factor by updating the hash
+      cy.visit(`/bouquets/${testTopic.id}#factor-${secondFactor.id}`)
+
+      cy.get(`#factor-${secondFactor.id}`).isInViewport({ wait: VIEWPORT_WAIT })
+      cy.contains(secondFactor.title).should('be.visible')
```
streino: This passes even if I comment out the 2nd `cy.visit`. Related to my 1st comment above.

src/views/topics/TopicDetailView.vue
streino: - Visit https://deploy-preview-893--ecospheres.sandbox.data.developpement-durable.gouv.fr/bouquets/test-clone-gr-1
- Go to Activité.
- Click on link "a ajouté le facteur "RPG" au bouquet (690f132ec0cf0e9efb0be01e)" => Sends you to Données on the corresponding factor ✔️. 
- Go back to Activité.
- Click again on link "a ajouté le facteur "RPG" au bouquet (690f132ec0cf0e9efb0be01e)" => Nothing happens ❌ (because #factor-xxx is still in the URL).

src/components/topics/TopicActivityList.vue
```diff
@@ -0,0 +1,90 @@
...
+  topic: Topic
+  factors: ResolvedFactor[]
+}>()
+
+const router = useRouter()
+const { pageConf } = useCurrentPageConf()
+
+const getActivityTranslation = (activity: Activity) => {
+  const elementId = getElementId(activity)
+  const activeFactor = getActiveFactor(elementId)
```
streino: Not sure about the use of "active" here. I initially thought it had something to do with selection, but it's about whether the factor still exists or not. Maybe the deep-linking stuff in this PR primed me to think "active == selected", but in any case maybe use something less ambiguous, or just `factor` = `getFactor(...)`?

src/components/topics/TopicActivityList.vue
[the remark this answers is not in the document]
abulte: Fixed. I'm now using events to navigate from Activity link instead of router + hash. We're going a bit more into Frankenstein territory, but we have shareable URLs (url based trigger), not too bad forward/backward handling (event based) and open in new tab (url based trigger).

cypress/e2e/custom/ecospheres/topics/elements/deeplink.cy.ts
abulte: Removed the test, not related to a use case.

Thesauruv:
@abulte @streino One small comment on the wording since our last call. I would replace "facteur" with "élément", which is more generic. The term "facteur" does not appear elsewhere in the interface.

image

abulte:
@Thesauruv done

streino (approved):

streino (approved):
Next time I'll wait for the tests to finish before reviewing 😅

PR #901, opened 2025-10-06
make the SolutionTag more resilient

JeSuisUnCaillou:
When we got from cas d'usage list to solution list, the card components change before the list of topics is updated, which triggers errors because solution topic's cards can't find its extra key in a cas d'usage topic's extras.

### DIFF
--- a/src/custom/simplifions/components/SimplifionsSolutionTag.vue
+++ b/src/custom/simplifions/components/SimplifionsSolutionTag.vue
@@ -29,7 +29,7 @@ const solution = computed(
 )
 
 const isPublic = computed(() => {
-  return solution.value.Public_ou_prive === 'Public'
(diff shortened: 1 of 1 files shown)


JeSuisUnCaillou:
Merci Sentry de m'avoir relevé ce bug :D

src/custom/simplifions/components/SimplifionsSolutionTag.vue
abulte, line 26: I'd type this to `| undefined` so that your `.value?` is enforced by the type system.

abulte: Yeah. Now, if you remove your fix, Typescript will warn you that `.value` can be `undefined`.

Capture d’écran 2025-10-06 à 14 56 31

PR #903, opened 2025-10-06
use DatagouvfrAPI version 2

JeSuisUnCaillou:
for improved performances on datasets with big resources like https://www.data.gouv.fr/datasets/5eebbc067a14b6fecc9c9976

The V2 API doesn't send all the data (we don't use) of the resources.


cypress/support/grist_mocks.js
```diff
@@ -27,24 +27,11 @@ Cypress.Commands.add('mockGristRecords', (resourceId, elements = []) => {
...
-    }
-  ).as(`get_grist_${resourceId}_record_${element.id}`)
-})
-
 Cypress.Commands.add('mockGristRecordsByIds', (resourceId, elements) => {
   const ids = elements.map((el) => el.id)
   cy.intercept(
     'GET',
-    `https://grist.numerique.gouv.fr/api/docs/*/tables/${resourceId}/records?filter=${JSON.stringify({ id: ids })}`,
+    `https://grist.numerique.gouv.fr/api/docs/*/tables/${resourceId}/records?filter=%7B%22id%22:%5B${ids.join(',')}%5D%7D`,
```
JeSuisUnCaillou, line 34: And this was not properly encoded, so I ended up putting the encoded character directly, even if it's not ideal to read.

PR #904, opened 2025-10-06
fix(indicators): reintroduce theme and enjeu as tags

abulte:
Fix https://github.com/ecolabdata/ecospheres/issues/811

#887 removed some indicators filters that were also used as tags on the detail page. Originally, filters were made to work hand-in-hand with tag display, but now we need to display some tags that aren't used as filters. 

This introduces an option `hide_on_list` for filters, with which we can prevent some filters from appearing on the list page, thus making them tag-only. It's a bit deceptive since those should not be filters at all, but it's not too complicated and keeps the meat of the tag system based on filter values working.

### DIFF
--- a/configs/simplifions/config.yaml
+++ b/configs/simplifions/config.yaml
@@ -252,11 +252,11 @@ pages:
         use_tag_prefix: true
         values:
           - id: proactivite
-            name: ⭐️⭐️⭐️ Proactivité | L'usager n'a plus de démarche à faire
+            name: 💠💠💠 Proactivité | L'usager n'a plus de démarche à faire
           - id: dlnuf
-            name: ⭐️⭐️ Dites-le nous une fois | L'usager n'a plus à fournir de justificatifs
+            name: 💠💠 Dites-le nous une fois | L'usager n'a plus à fournir de justificatifs
           - id: acces-facile
-            name: ⭐️ Accès facile | L'usager ou l'agent trouve l'information
+            name: 💠 Accès facile | L'usager ou l'agent trouve l'information
 
   # SOLUTIONS
   solutions:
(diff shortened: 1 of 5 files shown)


streino (commented):
I get the point about keeping the change simple but I'm worried about leaving extra complexity in the code -- "deceptive" 😬

What would separating filters and tags entail?

abulte:
Another solution would be to customize the indicators filters to manually exclude some of them. It would keep the modification scoped to indicators and preserve the generic filter logic. Would be more complex though, and _maybe_ we'll need the same feature on other objects later.

streino:
Could this just be a naming issue? We end up with "filters" that aren't really filters. 
Changing the `filters` section to something else and add a `filter` sub-section would make it more understandable IMO.

Config could look something like this:
```
  indicators:
    ...
    facets:  # Could be "tags" too, but it really is (poor person's) facets we're doing here
      - name: Thématique
        id: theme
        color: green-menthe
        use_tag_prefix: true  # Not sure where this one needs to go
        values: *fnv_themes
        filter: null
      - name: Enjeu
        id: enjeu
        color: blue-ecume
        use_tag_prefix: true
        values:
          - id: adaptation-climat
            name: Adaptation climat
          - id: attenuation-climat
            name: Atténuation climat
          - ...
        filter: null
      - name: Secteur
        id: secteur
        color: purple-glycine
        values:
          - id: agriculture-forets-sols
            name: Agriculture, Forêts et Sols
          - id: alimentation
            name: Alimentation
          - ...
        filter:
          type: select
          default_option: Tous les secteurs
```

abulte:
I tried... But it's too deep a refactoring (and I'm not even finishing, touching 10+ files and hundred of lines). It's not worth the slight inconvenience re naming.

abulte:
I'll add some quick testing though, will make future refactoring easier.

streino:
> I tried... But it's too deep a refactoring (and I'm not even finished, touching 10+ files and hundred of lines). It's not worth the slight inconvenience re naming.

Let's not block the fix and go with the low-impact solution then.

I still think it's worth refactoring later. Config is complex enough as it is, so let's try to minimize avoidable head-scratching :nail_care:

PR #906, opened 2025-10-09
feat(topic): open factor in QGIS

abulte:
Fix https://github.com/ecolabdata/ecospheres/issues/804

This generates a QLR file for a factor. Opening this file should trigger a QGIS launch plus load of the selected WMS or WFS stream.

~~It currently handles the first acceptable resource of a Dataset, which is the simplest solution. We could:~~
- ~~put all the streams for a dataset in the same QLR file — but this would duplicate the same stream in WFS and WMS for the majority of datasets~~
- ~~have a dropdown button that lists the available ressources for opening — not sure if we need that for a first version~~
- be a bit smarter and favor the first WFS, then WMS (probably the best one) -> done

~I'm not even sure we should care about WMS at all. Do we have datasets that only have WMS?~ False, we do have some lonely WMS, and it looks like the WFS is more useful (features infos)?

Open questions:
- ~~layer name is inferred from resource title, which probably will work on typical harvested datasets, but we should watch for exceptions~~ layer name is "validated" from resource title, and if it does not look viable, we fetch the WFS layers and open them all, or hide the button for WMS (harder to fetch layers + WFS has priority)
- current projection is EPSG:4326 (WGS 84), which is probably the most typical, but should we try to detect it somehow? I think there's an Inspire metadata for that, but I don't think we're harvesting it.
- i have no idea what some of the options in the XMLs do 😅 

Behavior:
  - Dataset with WFS + WMS → Opens WFS
  - Dataset with only WMS (valid layer) → Opens WMS
  - Dataset with only WMS (invalid layer) → Button hidden
  - Dataset with only WFS (no layer) → Opens all WFS layers

### DIFF
--- a/src/utils/qgis.ts
+++ b/src/utils/qgis.ts
@@ -55,7 +55,7 @@ function addMaplayer(
 
   maplayer.ele('id').txt(id)
   maplayer.ele('datasource').txt(datasource)
-  maplayer.ele('provider', { encoding: 'UTF-8' }).txt(provider)
+  maplayer.ele('provider').txt(provider)
   maplayer.ele('layername').txt(layerName)
 
   // Add CRS info from CRS_DEFINITIONS


abulte:
## WMS projection

What I understand about WMS projection handling:
- The server supports a list of projections, the first one being the "native" one. We can see them through a getCapabilities request
- QGIS will try to query the WMS with the current project projection
- We currently force EPSG:4326 when building the datasource. Once again, it's pretty standard these days.
- If we were to force an unsupported projection, it would probably fail (but my limited does not show that, I'm confused)

EPSG:4326 is probably a good default. Both streams I tested support it, and one of them does not (explicitly?) support EPSG:2154. Plus, it's a worldwide projection which works well from DROM-COM, whereas EPSG:2154 would not (AFAIK).

So we could:
- keep as is, forcing EPSG:4326
- remove the projection from the datasource URL. It works with the streams I'm tested. It should try with the first supported projection? Unclear, more tests needed.

https://services.sandre.eaufrance.fr/geo/zgr?request=GetCapabilities&service=wms

```
<CRS>EPSG:4326</CRS>
<CRS>EPSG:2154</CRS>
<CRS>EPSG:32620</CRS>
<CRS>EPSG:4559</CRS>
<CRS>EPSG:5490</CRS>
<CRS>EPSG:2972</CRS>
<CRS>EPSG:2975</CRS>
<CRS>EPSG:2989</CRS>
<CRS>EPSG:4471</CRS>
<CRS>EPSG:4467</CRS>
<CRS>EPSG:27572</CRS>
<CRS>EPSG:32622</CRS>
<CRS>EPSG:32738</CRS>
<CRS>EPSG:32740</CRS>
<CRS>EPSG:3857</CRS>
<CRS>EPSG:900913</CRS>
<CRS>EPSG:6749</CRS>
<CRS>EPSG:5490</CRS>
<CRS>EPSG:4258</CRS>
<CRS>EPSG:7409</CRS>
```

https://sextant.ifremer.fr/services/wms/oceanographie_physique?service=WMS&request=GetCapabilities&version=1.3.0

```
<CRS>CRS:84</CRS>
<CRS>EPSG:4326</CRS>
```

I'll keep default to 4326 for now.

Thesauruv:
@abulte Additional test on the RPG datasets (see in the bouquet) -> example of a confusing one. Among the resources listed:

- A WMTS service: https://demo.data.gouv.fr/datasets/rpg/#/resources/a43e03a8-4015-4818-8112-51c96050649a
- A WMS service: https://demo.data.gouv.fr/datasets/rpg/#/resources/56099928-e779-4545-b9cd-eb5f3ff8e8c7
- A WFS service: https://demo.data.gouv.fr/datasets/rpg/#/resources/d1bf6cf7-0f7e-4dee-a0d7-e54487e13b5a
- A WMS service: https://demo.data.gouv.fr/datasets/rpg/#/resources/8214a93d-49ae-4e1b-b5a1-f7c78394e743

I have the impression that the QGis export retrieves the second link ‘LANDUSE.AGRICULTURE.2007’ and not the WFS link, preventing the user to retrieve layer properties.

Furthermore at this stage, the difference between WMS and WMTS is not clear for me.

abulte:
Test on a WFS that supposedly does not support EPSG 4326 https://carto.smmar.fr/wxs?service=WFS&request=GetCapabilities -> stream opens correctly.

NB: the "Périmètre du SMMAR" dataset does not generate a correct QLR since the resource title is "WMS", which could be a valid layer name, but is not. Maybe we should:
- test if the detected layer exists -> QLR with only this layer
- if no detected layer or detected does not exist -> QLR with all the layers

[p_rim_tre_du_smmar.qlr.txt](https://github.com/user-attachments/files/23098161/p_rim_tre_du_smmar.qlr.txt)

src/utils/qgis.ts
```diff
@@ -0,0 +1,322 @@
...
+ * Each field corresponds to epsg.io API endpoints:
+ * - authid: EPSG code identifier (e.g. "EPSG:4326")
+ * - srid: Spatial Reference System ID - numeric part only (e.g. "4326")
+ * - proj4: Proj4 string from https://epsg.io/{code}.proj4
+ *
+ * To add a new CRS:
+ * 1. Fetch proj4 string from https://epsg.io/{code}.proj4
+ * 2. Add entry here with authid, srid, and proj4 fields
+ */
+const CRS_DEFINITIONS = {
```
streino: How about enum?

src/utils/ogcServices.ts
```diff
@@ -0,0 +1,206 @@
...
+  const parserError = doc.querySelector('parsererror')
+  if (parserError) {
+    throw new Error(`XML parsing failed: ${parserError.textContent}`)
+  }
+
+  return doc
+}
+
+/**
+ * Extracts base URL by removing GetCapabilities query parameters
```
streino: `service` is not limited to `request=GetCapabilities`, so "GetCapabilities parameters" is misleading (used here and in some tests). I'm surprised it works without the `service` param actually (should be "as required" as `version`).

src/utils/qgis.ts
```diff
@@ -0,0 +1,322 @@
...
+  const crsDef = CRS_DEFINITIONS[crs]
+
+  const maplayer = parent.ele('maplayer', {
+    type: type,
+    hasScaleBasedVisibilityFlag: '0'
+  })
+
+  maplayer.ele('id').txt(id)
+  maplayer.ele('datasource').txt(datasource)
+  maplayer.ele('provider', { encoding: 'UTF-8' }).txt(provider)
```
streino: Why the redundant encoding, and why only here?

src/utils/ogcServices.ts
```diff
@@ -0,0 +1,206 @@
...
+ * WFS URLs often come as: https://example.com/wfs?request=GetCapabilities&service=WFS
+ * This returns just the base URL: https://example.com/wfs
+ */
+export function extractBaseUrl(url: string): string {
```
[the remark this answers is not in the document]
streino: I sort of expect to be able to add path elements to a baseUrl, but that's not what it means here.

I don't have a better suggestion though, so I'm not going to insist on changing. Maybe update the example in the comment to make it clear there may be params left?

src/utils/qgis.ts
abulte: I'll leave it without. The defaults look sane on paper. A QGIS export does not have the attribute.

abulte: @Thesauruv I know you like looking at entities details in WFS :-) If you see some [mojibake](https://fr.wikipedia.org/wiki/Mojibake), eg `cafÃ©` instead of `café`, tell me.

streino (approved):

abulte:
- OCS GE : c'est un WMS, donc on ne vérifie pas le nom de la couche, et celui détecté via le titre de la ressource n'est pas bon "OCSGE.USAGE.2010". Si on faisait la même mécanique que pour les WFS, on ouvrirait toutes les couches du géoportail 😱. Est-ce qu'il a déjà marché ?
- RPG : ça marche chez moi, il ouvre "RPG.2010:rpg_2010". Tu as quelle erreur @Thesauruv ?

Capture d’écran 2025-12-11 à 10 22 06

Rien ne parait lié au refacto/review, donc je merge ici et on peut itérer plus tard sur des issues spécifiques.

PR #923, opened 2025-10-29
fix: escape ci comment

abulte:
Writing a comment with `xxx` and a /deploy somewhere would execute the xxx command 😬 

Eg https://github.com/opendatateam/udata-front-kit/pull/830#issuecomment-3460461687


abulte:
`cat /etc/passwd`

PR #936, opened 2025-11-03
feat: filter bouquets by organization

abulte:
Fix https://github.com/ecolabdata/ecospheres/issues/80

This also makes the organization filter generic, no more `filtersComponent` override.


streino (commented):
Looks good but I don't see orgs for draft bouquets? I thought they were to be [included](https://github.com/ecolabdata/ecospheres/issues/80#issuecomment-3422275619)?

streino (approved):

PR #946, opened 2025-11-06
chore: remove dead query.drafts code

abulte:


abulte:
@streino not the query manipulation you're waiting for 😇

streino (approved):
> @streino not the query manipulation you're waiting for 😇

I would have been surprised if it was! Talk about complexity...

PR #954, opened 2025-11-12
fix robots workflow for other verticals

JeSuisUnCaillou:


abulte (commented):
Shouldn't the solution be to have a correct `.env.(smtg)` content on the target branch?

PR #983, opened 2025-11-26
chore: add .npmrc with ignore-scripts

abulte:
https://mattermost.incubateur.net/betagouv/pl/fia9offq37gazdcnw5wuniotxh


abulte:
Difficult with cypress and prepare (husky). I'll probably try to switch to pnpm instead.

PR #1013, opened 2025-12-23
feat(topics): open topic in QGIS

abulte:
Fix https://github.com/ecolabdata/ecospheres/issues/805

Adds support for topic-level QGIS export and heavily refactors the single factor export from #906.

Careful crafted example topic: https://deploy-preview-1013--ecospheres.sandbox.data.developpement-durable.gouv.fr/bouquets/alb-qgis-multi-factors

The structure now looks like this (first top level group is a topic, second one is a factor):

Capture d’écran 2025-12-24 à 15 40 10

It's a bit noisy but it's easier to understand what's been extracted from the Topic > resource chain, in the code and in QGIS. Plus, many exports can live alongside in QGIS while staying tidy. There's none in the example Topic, but if a WFS does not provide a valid layer name and we fetch all the layers, they will be grouped under "Resource title".

Capture d’écran 2025-12-25 à 15 53 46

### DIFF
--- a/src/components/topics/TopicFactorsList.vue
+++ b/src/components/topics/TopicFactorsList.vue
@@ -410,11 +410,13 @@ defineExpose({
                     target="_blank"
                     >Accéder au catalogue</a
                   >
-                  <button
+                  <DsfrButton
                     v-if="
                       factor.element?.id && ogcLayerInfo.has(factor.element.id)
                     "
-                    class="fr-btn fr-btn--sm fr-btn--secondary inline-flex"
+                    secondary
+                    size="sm"
+                    icon="fr-icon-road-map-line"
                     @click="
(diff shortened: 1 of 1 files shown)


streino (dismissed):

abulte:
@Thesauruv I unchecked only the "final" layer, because QGIS doesn't (by default) check in cascade. Let me know what you think.

abulte:
> Minor comment, could you include the "map" pictogram in the QGis export button at factor level, as in the "bouquet" export one?

👍 

> I anticipate some frustration among users who may expect the entire bouquet to be exported.

I'd say we can iterate on that when we have IRL feedack ? Maybe a modal that explains what's going on, or an info affordance somewhere (but difficult with a button).

PR #1018, opened 2026-01-12
chore: remove datasetEditorialization config

abulte:
No site uses `datasetEditorialization: false` anymore. This PR removes the setting and the associated (dead) code.

NB: simplifions bypasses all that (list and creation form), so it's not affected.


streino (approved):
🧹

PR #1029, opened 2026-01-16
fix(topics): canAdd(To)Topic checks for create route

abulte:
Fix https://linear.app/pole-api/issue/API2-516/erreur-quand-connecte-la-route-de-creation-dobjet-nexiste-pas

A site can use Topic and oauth without a creation form (eg simplifions). This make `canAdd(To)Topic` aware of the creation form route existence.

### DIFF
--- a/src/views/topics/TopicsListView.vue
+++ b/src/views/topics/TopicsListView.vue
@@ -25,7 +25,7 @@ const props = withDefaults(defineProps<Props>(), {
 
 const router = useRouter()
 const route = useRoute()
-const { meta, pageConf } = useCurrentPageConf()
+const { meta, pageConf, pageKey } = useCurrentPageConf()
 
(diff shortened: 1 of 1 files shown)


src/views/datasets/DatasetDetailView.vue
```diff
@@ -31,27 +30,30 @@ const datasetId = route.params.item_id
...
+const topicPageConf = topicPageKey ? usePageConf(topicPageKey) : null
 
 const canEdit = computed(() => {
   return pageConf.editable && userStore.hasEditPermissions(dataset.value)
 })
 const canAddToTopic = computed(() => {
-  return topicPageConf && userStore.loggedIn && canAddTopic
+  return (
+    topicPageConf &&
+    topicPageKey &&
```
streino: Test on `topicPageKey` is redundant, since `topicPageConf` is determined by it.

PR #1116, opened 2026-03-13
feat(simplifions): solutions integratices by category

abulte:
Fix https://linear.app/pole-api/issue/API-6465/ajouter-les-onglets-par-categories-de-solution-integratrices-dans-la

Uses `Solution.liste_categories_de_solution` to display solutions intégratrices in tabs.

➕ fix a bug where cas d'usages in filter list where not necessarily linked to visible solutions (and thus yield no result)

Capture d’écran 2026-03-13 à 11 19 13

### DIFF
--- a/src/custom/simplifions/components/SimplifionsSolutionIntegrateursTabs.vue
+++ b/src/custom/simplifions/components/SimplifionsSolutionIntegrateursTabs.vue
@@ -11,7 +11,9 @@
       tab-id="tab-logiciel-metier"
       class="fr-tabs__panel--white"
     >
-      <ul class="fr-grid-row fr-grid-row--gutters list-none fr-tabs__panel--beige-gris-galet">
+      <ul
+        class="fr-grid-row fr-grid-row--gutters list-none list--beige-gris-galet"
+      >
         <li
           v-for="integrateur in filteredLogicielsMetiers"
           :key="integrateur.id"
@@ -36,7 +38,9 @@
       tab-id="tab-brique-technique"
       class="fr-tabs__panel--white"
     >
-      <ul class="fr-grid-row fr-grid-row--gutters list-none fr-tabs__panel--beige-gris-galet" >
+      <ul
+        class="fr-grid-row fr-grid-row--gutters list-none list--beige-gris-galet"
+      >
         <li
           v-for="integrateur in filteredBriquesTechniques"
           :key="integrateur.id"
@@ -61,7 +65,9 @@
       tab-id="tab-portail-consultation"
       class="fr-tabs__panel--white"
     >
-      <ul class="fr-grid-row fr-grid-row--gutters list-none fr-tabs__panel--beige-gris-galet">
+      <ul
+        class="fr-grid-row fr-grid-row--gutters list-none list--beige-gris-galet"
+      >
         <li
           v-for="integrateur in filteredPortailsConsultation"
           :key="integrateur.id"
@@ -154,10 +160,7 @@ watch(tabTitles, () => {
   background-color: white;
 }
 
-.fr-tabs__panel--beige-gris-galet {
+.list--beige-gris-galet {
   background-color: var(--background-alt-beige-gris-galet);
 }
-
-
-
 </style>


JeSuisUnCaillou (commented):
En regardant par moi-même, le seul truc qui a attiré mon attention c'est l'absence de tests sur ces onglets.

src/custom/simplifions/components/SimplifionsSolutionIntegrateursTabs.vue
```diff
@@ -0,0 +1,145 @@
...
+}>()
+
+const activeTab = ref(0)
+
+const filteredLogicielsMetiers = computed(() =>
+  props.solutions.filter((sol) =>
+    sol.fields.liste_categories_de_solution?.includes('Logiciel métier')
+  )
+)
+
```
JeSuisUnCaillou: Remove the `💠💠💠` placeholder — this appears on "Logiciels métiers" and "Briques techniques" but not on "Portails de consultation", making it clearly inconsistent. These look like debug markers that were forgotten.

src/custom/simplifions/components/SimplifionsSolutionIntegrateursTabs.vue
```diff
@@ -0,0 +1,145 @@
...
+  <DsfrTabs
+    v-model="activeTab"
+    tab-list-name="Catégories de solutions intégratrices"
+    :tab-titles="tabTitles"
+  >
+    <DsfrTabContent
+      v-if="filteredLogicielsMetiers.length"
+      panel-id="tab-content-logiciel-metier"
+      tab-id="tab-logiciel-metier"
+      style="background-color: white"
```
JeSuisUnCaillou: Using `v-if` on `DsfrTabContent` while `tabTitles` is a computed that dynamically skips empty categories is fragile — `DsfrTabs` maps panels by DOM order, so any mismatch between rendered panels and tab titles could break the active tab. Prefer `v-show` to keep all three panels in the DOM.

src/custom/simplifions/model/grist.ts
```diff
@@ -110,6 +110,7 @@ export type Solution = {
   Type_de_solution: string[]
   URL_demande_d_acces: string
   solutions_integratrices: number[] | null
+  liste_categories_de_solution: string[]
```
JeSuisUnCaillou: The field is declared as `string[]` (non-optional) but the computed filters in the new component use `?.includes()`, implying it can be `undefined`. Should be `string[] | undefined` to match reality.

src/custom/simplifions/components/SimplifionsSolutionIntegrateursTabs.vue
```diff
@@ -0,0 +1,145 @@
...
+    >
+      <ul class="fr-grid-row fr-grid-row--gutters list-none">
```
JeSuisUnCaillou: Inline `style` should be replaced with a DSFR utility class or a scoped CSS rule.

JeSuisUnCaillou (commented):
En faisant faire la review à Claude, il m'a trouvé ça :point_down: 

## Code review

### 🔴 Blocker — Emoji placeholders left in tab titles

~~Two tab titles contain `💠💠💠` markers that look like debug/placeholder leftovers. The third tab ("Portails de consultation") does not have them, which makes this clearly inconsistent. These should be removed before merging.~~

Irrelevant

---

### 🟡 Risk — `v-if` on `DsfrTabContent` may break tab index mapping

The three `DsfrTabContent` blocks use `v-if` to skip empty categories. At the same time, `tabTitles` is a computed that also skips them. While logically consistent, DSFR's `DsfrTabs` maps tabs to panels by DOM order/index. If the first category is empty, `activeTab = 0` will point to the first rendered panel (e.g. "Briques techniques") while the user sees tab 0 as "Briques techniques" — this should be fine actually — but it's fragile. A safer pattern is to always render all panels with `v-show` to keep the DOM structure stable and predictable.

---

### 🟡 Type inconsistency — `liste_categories_de_solution` should be optional

The field is declared as `string[]` (non-optional) in the type, but the computed filters use optional chaining (`?.includes(...)`), implying it can be absent. The type should be `string[] | undefined` to match reality.

---

### 🟡 Style — inline `style="background-color: white"` on all three panels

Should use a DSFR utility class or a scoped CSS rule instead of inline styles.

---

### 🟢 Minor — no new E2E test for tab behavior

The Cypress fixture is updated, but no new test cases cover tab navigation or per-category filtering. Low risk given the simplicity, but worth adding.

JeSuisUnCaillou (dismissed):

abulte:
> > est-ce que tu peux utiliser une classe scopée pour ça ? cf fr-tabs__panel--white dans le même fichier.
> 
> Je viens d'envoyer un correctif, j'espère que c'était ce qui était attendu

J'ai changé le nom, mais sinon c'est bon ! Je veux bien un dernier approve avant de merger.

PR #1128, opened 2026-03-17
feat(dataservice): add stats box

abulte:
Fix https://github.com/ecolabdata/ecospheres/issues/973

Capture d’écran 2026-03-17 à 16 31 43

### DIFF
--- a/src/components/MetricsStatBoxes.vue
+++ b/src/components/MetricsStatBoxes.vue
@@ -17,10 +17,16 @@ const metrics = ref<DatasetMetrics | DataserviceMetrics | null>(null)
 watchEffect(async () => {
   if (!props.objectId) return
   try {
-    metrics.value =
-      props.objectType === 'dataset'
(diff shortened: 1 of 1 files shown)


src/views/dataservices/DataserviceDetailView.vue
```diff
@@ -203,6 +203,11 @@ onMounted(() => {
             >Faire une demande d'habilitation</a
           >
         </div>
+        <!-- metrics -->
+        <MetricsStatBoxes
+          object-type="dataservice"
+          :object-id="dataservice.id"
+        />
```
streino: Feels this should be a oneliner. [It is for datasets](https://github.com/opendatateam/udata-front-kit/pull/1128/changes#diff-b6280e7c6d2d49b5637d49465ab8308b06c759dac4bd703e1f33897c11bbce25R94). And if working as intended, maybe we should increase line width a bit?

streino (approved):

PR #1153, opened 2026-04-07
feat: unified search

### LINKED ISSUE #993: Recherche unifiée
### Contexte

Réimplémentation ou intégration du composant de recherche unifiée pour ecologie.data.gouv.fr, en s'appuyant sur le composant data.gouv.fr comme base, avec des adaptations spécifiques à notre univers.

---

### Types d'objets à inclure

| Type | Statut | Notes |
|---|---|---|
| Jeux de données | ✅ | |
| Indicateurs | ✅ | À paramétrer dans le composant |
| Collections | ✅ | À paramétrer dans le composant |
| API | ✅ | |
| Organisations | ❌ hors scope v1 | Dépendance au fichier JSON exporté, pas requêtable nativement via la search stack |
| Réutilisations | ❌ hors scope v1 | Nécessite de construire un univers de réutilisation d'abord |

---

### Filtres mis en avant (sous la barre de recherche)

Par rapport au composant data.gouv.fr, on **supprime** :

- ❌ **Format de données** — métadonnées non fiables en full moissonné, trop de faux positifs
- ❌ **Modalité d'accès** — non mappé au moissonnage, zéro résultat sur ecologie.data.gouv.fr
- ❌ **Type d'organisation** — 99% service public dans notre univers, filtre sans valeur ajoutée

On **garde** :

- ✅ **Date de mise à jour** (`last_update_range`) — utile, logique déjà implémentée côté data.gouv.fr
- ✅ **Label de données** (dont "Service public de données de référence") — même forme que le composant data.gouv.fr

---

### Filtres avancés

Filtres spécifiques à notre univers, à afficher **sans les grouper** pour éviter de les invisibiliser :

- ✅ **Organisation** — filtre custom, incontournable pour notre public
- ✅ **Thème Inspire** — custom, présentation en liste (trop de valeurs pour une facette standard)
- ✅ **Maille géographique** — custom, peu de valeurs → envisager un affichage en facette

---

### Tri

Options conservées : pertinence, date de création, date de modification, titre.

- ❌ Nombre d'abonnés / nombre de réutilisations : pas d'effort particulier pour l'instant

---

### Comportement de navigation entre onglets

- ✅ La query persiste quand on change d'onglet
- ✅ Si un filtre actif n'est pas compatible avec le nouvel onglet, il est silencieusement ignoré et la recherche se relance sans lui
- ❌ Ne pas reproduire la page éditorialisée intercalée dans l'onglet "données" de data.gouv.fr (étape irritante, hors scope du composant)

---

### Hypothèses techniques

- Base : composant data.gouv.fr, paramétré au maximum plutôt que réécrit
- Le composant sera monté sur toutes les routes concernées (même instance) — à confirmer à l'implémentation
- La recherche par organisation reste custom (pas de requête native via la search stack)

---

### Hors scope — tickets séparés à créer

- Univers réutilisations
- Univers organisations (indexation dans la search stack)
- Filtre "prévisualisation disponible" (nécessite un extra Hydra sur le jeu de données)
- Filtre modalité d'accès (une fois le mapping moissonnage corrigé)

--- 

Transcript complet réunion : https://docs.numerique.gouv.fr/docs/0b500c3c-8e0b-4b17-bc8b-5f024b73d721/?utm_source=docssharelink&utm_campaign=0b500c3c-8e0b-4b17-bc8b-5f024b73d721

abulte:
Fix https://github.com/ecolabdata/ecospheres/issues/993

Integrates `GlobalSearch` component from `@datagouv/components-next` that replaces all list pages from the front-kit and brings the "unified" search style of data.gouv.fr.

Capture d’écran 2026-05-04 à 16 07 15


abulte:
https://github.com/opendatateam/udata-front-kit/pull/1153/commits/a0d3f30318e34a694d1642be90ed02a6c2e314ed

## What was done

Replaced the DataList paradigm (per-page Pinia stores, `DatasetList`, `DataservicesList`, `PageFilters`, `SearchAPI`, filter utilities) with `GlobalSearch` from `@datagouv/components-next`.

### Architecture

- **`UnifiedSearchView.vue`** — single list view for all page types.
- **`useGlobalSearchPageRoutes()`** — replaces `useDatasetSearchPageRoutes`, `useDataserviceSearchPageRoutes`, `useTopicSearchPageRoutes`. Reads `object_type`, `universe_query`, and `filters[].search_display` from YAML config; only component references are passed in JS.
- **`object_type`** field in YAML — declares the page's object type (`datasets`, `dataservices`, `topics`). Single source of truth.
- **`search_display`** field on filter definitions — `basic` or `advanced`. Native GlobalSearch filter types (e.g. `last_update_range`, `badge`, `organization`) are passed directly to GlobalSearch config. Custom predefined-value filters (`type: select`) are rendered via `SearchSelectFilter.vue` in the `#filters` slot.
- **`SearchSelectFilter.vue`** — generic config-driven select filter, bound to `api_param ?? id` URL query param. 
- All 8 sites migrated to `useGlobalSearchPageRoutes`.

### Topics (bouquets/collections)

`GlobalSearch` does not support `topics` as a search class upstream. Topics pages use `SearchComponent` + `TopicList` (existing components) as an interim solution.
Filters (`PageFilters`) are not shown for now pending upstream support.
Tracked in [ecolabdata/ecospheres#1010](https://github.com/ecolabdata/ecospheres/issues/1010).

## What's next

### delete old infrastructure (after validation)

- `src/views/data/DataListView.vue`
- `src/views/topics/TopicsListView.vue`
- `src/components/datasets/DatasetList.vue`
- `src/components/dataservices/DataservicesList.vue`
- `src/components/pages/PageFilters.vue`
- `src/components/SearchComponent.vue`
- `src/store/GenericSearchStore.ts`, `DatasetSearchStore.ts`, `DataserviceSearchStore.ts`
- `src/services/api/SearchAPI.ts`, `DataserviceSearchAPI.ts`
- `src/utils/filters.ts`

### upstream

- Topics support in `GlobalSearch` ([#1010](https://github.com/ecolabdata/ecospheres/issues/1010)) → remove `SearchComponent`+`TopicList` fallback, restore filters for topics pages
- Tag-merging in `useStableQueryParams` → unblock theme/enjeu/secteur filters for indicators

abulte:
https://github.com/opendatateam/udata-front-kit/pull/1153/commits/3b73a678eaba47cb505fba3eed3933fe7830ec90

## What was done

Integrated GlobalSearch topic support, now available upstream (`@datagouv/components-next` ≥ 1.0.2-dev.40).

- **Topics fully handled by `GlobalSearch`** — `UnifiedSearchView` now uses `GlobalSearch` with `class: 'topics'` and a `#topic` slot for all topic pages. The interim `SearchComponent` + `TopicList` fallback is removed.
- **`buildGlobalSearchConfig` uses typed defaults** — `getDefaultTopicConfig` / `getDefaultDatasetConfig` / `getDefaultDataserviceConfig` are used with YAML-driven overrides, so `sortOptions` and other required fields are always provided.
- **Filters for topics** — `last_update_range` (basic) and `organization` (advanced) configured via `search_display` in YAML for the bouquets page.
- **Fix: `topicConf` no longer leaks to list route props** — was causing a Vue extraneous non-props warning on `UnifiedSearchView`.

Closes [ecolabdata/ecospheres#1010](https://github.com/ecolabdata/ecospheres/issues/1010).

## What's next

### delete old infrastructure (after validation)

- `src/views/data/DataListView.vue`
- `src/views/topics/TopicsListView.vue`
- `src/components/datasets/DatasetList.vue`
- `src/components/dataservices/DataservicesList.vue`
- `src/components/pages/PageFilters.vue`
- `src/components/SearchComponent.vue`
- `src/store/GenericSearchStore.ts`, `DatasetSearchStore.ts`, `DataserviceSearchStore.ts`
- `src/services/api/SearchAPI.ts`, `DataserviceSearchAPI.ts`
- `src/utils/filters.ts`

### upstream

- Tag-merging in `useStableQueryParams` → unblock theme/enjeu/secteur filters for indicators

abulte:
## What was done

Upstream changes: https://github.com/datagouv/cdata/pull/1034

### tagFilters for indicator and bouquet filters

- **`buildGlobalSearchConfig`** now derives `tagFilters` from `select`-type filters with `search_display` set and no `api_param` (or `api_param: tag`). Tag values follow the existing convention: `{filter_prefix}-{filter.id}-{value.id}` when `use_filter_prefix: true`.
- Indicators: `theme`, `enjeu`, `secteur`, `levier`, `maille` now appear as `tagFilters` (added `search_display: basic` in YAML).
- Bouquets: `Thématique` likewise (added `search_display: basic`).
- `datasets.inspire` filter migrated from the `#filters` slot to `tagFilters` (removed redundant `api_param: tag` from YAML).
- `SearchSelectFilter.vue` deleted — no longer needed.

### UnifiedSearchView migration gaps closed

- **Page title (h1)** restored — was missing after the DataListView/TopicsListView removal.
- **"Add collection" button** — shown to authenticated users with `canAddTopic` permission, matching the old `TopicsListView` behaviour.
- **`organization-url`** passed to `DatasetCard` and `TopicCard` slots.
- **Local `TopicCard`** is now the default for all topic pages (`TopicCard` from `@datagouv/components-next` was silently used after the migration, losing ecosphères-specific card content). Added `topicUrl` and `organizationUrl` props to the local card.
- **`list_all` guard** — `onMounted` redirect to `not_found` if `pageConf.list_all` is false, matching old behaviour.

### Phase 2 deletions

- `DataListView.vue`, `TopicsListView.vue`
- `DatasetList.vue`, `DataservicesList.vue`
- `PageFilters.vue`
- `GenericSearchStore.ts`, `DatasetSearchStore.ts`, `DataserviceSearchStore.ts`

Remaining live: `SearchComponent.vue`, `SearchAPI.ts`, `utils/filters.ts` — still used outside the list page context (header search, topic forms).

## What's next

- **UI iteration** styles from upstream `@datagouv/components-next` components (GlobalSearch sidebar, tagFilters selects, etc.) are not all rendering correctly in the local theme; needs CSS alignment work.
- **Organisations** port the custom organisation filter to the new component (need another fix upstream?)
- **Type facets** object types should be available for navigation in search facets. 
- **Tests** fix and migrate existing tests
- **Dependency** use dev version of https://github.com/datagouv/cdata/pull/1034 when merged and published
- Clean up `SearchAPI.ts`, `utils/filters.ts`, `SearchComponent.vue` once their remaining usages are migrated (topic forms, header search — separate scope).

abulte:
## What was done

### CSS fix: `.input` class missing from `components.css`

The search bar (narrow, no background) and the SearchableSelect components (collapsed/unstyled) were broken because `.input` — the base input class used in `SearchInput.vue` and `SearchableSelect.vue` — was defined in the upstream's Tailwind app build but not exported in the `components.css` distributable.

Fixed upstream in datagouv/cdata#1034 ([commit `4804e6f`](https://github.com/datagouv/cdata/commit/4804e6f21f6020ab3822620bc3699e0f2234d792)): added `.input` to `assets/main.css` so it lands in the built `components.css`.

## What's next

- **UI iteration** remaining style gaps after the `.input` fix (GlobalSearch sidebar layout, spacing, etc.)
- **Organisations** port the custom organisation filter to the new component (may need another upstream fix)
- **Type facets** object types should be available for navigation in search facets
- **Tests** fix and migrate existing tests
- **Dependency** use stable release of datagouv/cdata#1034 when merged and published
- Clean up `SearchAPI.ts`, `utils/filters.ts`, `SearchComponent.vue` once their remaining usages are migrated (topic forms, header search — separate scope)

abulte:
## What was done

### CSS fix: `.input` class missing from `components.css`

(see [previous status comment](https://github.com/opendatateam/udata-front-kit/pull/1153#issuecomment-4224210212))

Fixed upstream in datagouv/cdata#1034 — waiting for merge and stable release.

### Organisations: org filter implemented via facets

The organisation filter is now implemented using the facets approach — ecolabdata/cdata#1.

## What's next

- **Dependency** waiting for datagouv/cdata#1034 to be merged and published (unblocks the `.input` CSS fix)
- **Tag filters** waiting for `tagFilters` to land upstream to propose ecolabdata/cdata#1
- **Type facets** object types filter needs upstream changes — ecolabdata/ecospheres#1021
- **Spec resync** need to reconcile original issue specs vs current filter order / placement / existence
- **Tests** fix and migrate existing tests
- Clean up `SearchAPI.ts`, `utils/filters.ts`, `SearchComponent.vue` once remaining usages are migrated (topic forms, header search — separate scope)

abulte:
## What was done

### Upstream dependencies resolved

- [datagouv/cdata#1048](https://github.com/datagouv/cdata/pull/1048) (custom filters via `useSearchFilter`) merged — replaces the facets-based org filter approach. `SearchSelectFilter` now uses the upstream composable, and `SearchOrganizationFilter` follows the same pattern.
- `.input` CSS fix landed in [datagouv/cdata#1034](https://github.com/datagouv/cdata/pull/1034).

### Type selector

- Sidebar radio group implemented in `GlobalSearch` for switching between search types (datasets, indicators, collections, API).
- Query persists across type switches; incompatible filters are silently dropped and search re-runs.
- Sort options reset when switching to a type that does not support the current sort value.
- Scroll position preserved during type transitions.

### Sort options

- Dataset sort options explicitly limited to `Date de création` and `Dernière mise à jour` (removed `Nombre d\abonnés` / `Nombre de réutilisations` per [issue #993 spec](https://github.com/ecolabdata/ecospheres/issues/993)).

### Organization filter

- Custom organization filter resurrected from legacy code (`OrganizationStore` + JSON config files) and integrated as a `CustomFilterConfig` alongside select filters.
- Org list refreshes when switching types (watched `pageKey` prop).
- `type: organization_facet` → `type: organization` in YAML config for consistency.

### Unified filter config architecture

- `tagFilters` + `organizationFilter` merged into a single `customSelectFilters: CustomFilterConfig[]` array.
- `CustomFilterConfig` is a discriminated union: `CustomSelectFilterConfig` (static values) | `OrganizationFilterConfig` (dynamic JSON loading via `pageKey`).
- Single `v-for` loop in `UnifiedSearchView` renders the appropriate component based on the union discriminant (`'values' in filter`).
- `f.search_display` guard factored out in the filter builder.

## What's next

- **Spec resync** — reconcile remaining items from [issue #993](https://github.com/ecolabdata/ecospheres/issues/993):
  - Organisation placement: spec says advanced filter, currently basic on datasets/dataservices/bouquets
  - Maille géographique: spec suggests facet display (few values), currently a select dropdown
  - Verify all removed filters (format, access_type, producer_type) are truly absent from config
- **UI iteration** — upstream component styles (GlobalSearch sidebar, filter spacing, etc.) need CSS alignment with ecosphères theme
- **Tests** — fix and migrate existing tests for the new architecture
- **Cleanup** — `SearchAPI.ts`, `utils/filters.ts`, `SearchComponent.vue` still used outside list page context (topic forms, header search) — separate scope

abulte:
## What was done

### Config & architecture

- `type: organization` renamed to `type: organization_custom` across all YAML configs (ecospheres, culture) and the type model — avoids collision with the native upstream filter type, which now uses the same `organization` key.
- `search_display` made mandatory for filters to appear in `GlobalSearch`: missing `search_display` now emits a `console.warn` instead of silently skipping (simplifions select filters were all missing it and are now fixed).
- `customSelectFilters` builder refactored for clarity (explicit `isRenderable` predicate).

### Component cleanup

- Removed local `NoResults.vue` and `TopicList.vue` (both superseded by upstream `GlobalSearch`).
- Removed `cardClass`, `listComponent`, `filtersComponent` from `SearchPageRoutesOptions` and `RouteMeta` — no longer needed.
- Removed legacy `DataservicesAPI.search()` override (v1 endpoint workaround no longer required).
- Switched from yalc local link to published `@datagouv/components-next@1.0.2-dev.61`.

### UI

- `<h2>` added above the search input in `UnifiedSearchView`, sourced from `pageConf.search.input`.
- Removed `organization-url` prop from topic card in `UnifiedSearchView` (org link on topic cards dropped).
- Types order fixed in `ecospheres/config.yaml` (dataservices and bouquets sections moved to correct position).

### Tests

- `datagouv_mocks.js` page size updated 10 → 20 to match upstream default.
- `expectActionToCallApi` helper rewritten: now takes a `Record<string, string | string[]>` and does order-independent URL param matching — more robust for multi-value params.
- Simplifions, ecospheres (datasets/dataservices/topics), and culture tests updated for new page size, pagination selectors, and filter slot structure.
- Obsolete logistique test removed.

### Known debt (documented in code)

- **FIXME in `router/utils.ts`**: `DatasetFilterKey` / `DataserviceFilterKey` are inferred from upstream return shapes (`ReturnType<...>['basicFilters'][number]`) because `@datagouv/components-next` does not export the filter interfaces. Will silently break if upstream renames `basicFilters`. Ask upstream to export the types.

## What's next

- **Spec resync** — reconcile remaining items from [issue #993](https://github.com/ecolabdata/ecospheres/issues/993):
  - Organisation placement: spec says _advanced_ filter, currently `basic` on datasets/dataservices/bouquets
  - Maille géographique: spec suggests facet display (few values), currently a select dropdown
  - Verify all removed filters (`format`, `access_type`, `producer_type`) are truly absent from configs
- **UI iteration** — GlobalSearch sidebar, filter spacing, and overall layout still need CSS alignment with the écosphères theme
- **Cleanup** — `SearchAPI.ts`, `utils/filters.ts`, `SearchComponent.vue` remain in use outside list page context (topic forms, header search) — out of scope for this PR, needs a follow-up
- **Cleanup** : private, spatial_zone, spatial_granularity, and checkbox are legacy/not-yet-supported filters, should remove?

abulte:
## What was done

### Bug fixes

- **Dataservice indicator card** — `DatasetOrIndicatorCard` was mistakenly registered as `cardComponent` on the `dataservices` route, causing it to receive `Dataservice` objects instead of `DatasetV2` in the list view. Fixed by using `datasetCardComponent` instead (scoped to the datasets tab in the detail view), leaving `cardComponent` unset so the list falls back to `DataserviceCard`.
- **Contrast tests** — selector updated from `ul.fr-grid-row > li` (old list DOM) to `p[role="status"]` (GlobalSearch result count, only rendered when results are loaded).

### Config: `search.placeholder`

- Added `pages.xxx.search.placeholder: string | null` to `PageSearchConf` and wired it into the `GlobalSearch` config for all three search types via `buildSingleTypeConfig`.
- `null` = suppress the upstream default placeholder (empty input); omitting the key lets the upstream default through. A conditional spread is used to avoid accidentally including `placeholder: undefined` in the config object, which would trigger GlobalSearch's `'placeholder' in cfg` check and silently suppress the default.
- All site configs set `placeholder: null`.

### Legacy filter cleanup

Four filter types were audited and cleaned up:

- **`spatial_granularity`** — dead (doc comments only, never in any config, no handling logic). Removed from `PageFilterConf` type union.
- **`spatial_zone`** — misnamed and dead in configs (accessibilite, logistique), with no handling logic. Removed from the type union; the two config entries now use the correct native `type: geozone` with `search_display: basic`.
- **`checkbox`** — zero config usage, no rendering path, `CheckboxComponent.vue` never imported anywhere. Removed from the type union, from `useFiltersState`, and `CheckboxComponent.vue` deleted.
- **`private`** ("Afficher les brouillons") — silently broken: GlobalSearch bypasses `TopicStore` entirely, so the toggle had no effect on API calls. Removed from all configs (ecospheres bouquets, simplifions ×2), from the type union, from `useFiltersState`. `useCheckboxQuery` deleted (only handled `private`). `TopicStore.query()` deleted (no callers, sole user of `useCheckboxQuery`). `usePageQueryParams` simplified accordingly. The `private` API param is also not supported by the GlobalSearch backend.

Additional config fixes:
- Logistique bouquets `theme` filter was missing `search_display: basic` — fixed (was emitting a console warning on every page load).

### Header search dropdown removal

The `dropdown` header search feature (Multiselect routing to specific list pages) was only actively used by simplifions. Removed from: simplifions and ecospheres configs, `HeaderSearchConf` model, `HeaderComponent`, and `SearchComponent` (now a simpler component, `@vueform/multiselect` dependency dropped from this path).

### Tests

- Culture filter tests rewritten using `cy.selectFilterValue()` and `cy.expectActionToCallApi()` (old tests used `.fr-select-group` / `.multiselect-wrapper` selectors from the pre-GlobalSearch DOM). Switched to `cy.mockListApis()`.
- Culture filter `search_display: basic` added to all three filters (tag, organization_badge, organization_custom) — they were missing it and silently not rendering. Skipped test block unskipped.
- Simplifions navigation tests: two tests that relied on the removed dropdown routing removed.

## What's next

- **Spec resync** — reconcile remaining items from [issue #993](https://github.com/ecolabdata/ecospheres/issues/993):
  - Organisation placement: spec says _advanced_ filter, currently `basic` on datasets/dataservices/bouquets
  - Maille géographique: spec suggests facet display (few values), currently a select dropdown
  - Verify all removed filters (`format`, `access_type`, `producer_type`) are truly absent from configs
- **UI iteration** — GlobalSearch sidebar, filter spacing, and overall layout still need CSS alignment with the écosphères theme
- **Cleanup** — `SearchAPI.ts` and `SearchComponent.vue` remain in use outside list page context (topic forms, header search) — out of scope for this PR, needs a follow-up. (`utils/filters.ts` is now cleaned up.)
- **FIXME** — `DatasetFilterKey` / `DataserviceFilterKey` are inferred from upstream return shapes because `@datagouv/components-next` does not export the filter interfaces. Ask upstream to export them.

configs/ecospheres/config.yaml
```diff
@@ -292,14 +293,32 @@ pages:
         display: true
     editable: true
     filters:
+      - name: Mise à jour
+        id: last_update_range
+        type: last_update_range
+        search_display: basic
+        values: []
```
abulte: Remove

configs/ecospheres/config.yaml
```diff
@@ -292,14 +293,32 @@ pages:
...
+      - name: Label
+        id: badge
+        type: badge
+        search_display: basic
+        values: []
```
abulte: Remove

configs/ecospheres/config.yaml
```diff
@@ -292,14 +293,32 @@ pages:
...
         id: organization
-        type: organization
+        type: organization_custom
+        search_display: basic
+        values: []
+      - name: Granularité
+        id: granularity
+        type: granularity
+        search_display: advanced
+        values: []
```
abulte: Remove

configs/ecospheres/config.yaml
```diff
@@ -292,14 +293,32 @@ pages:
...
+        type: badge
+        search_display: basic
+        values: []
       - name: Organisation
         default_option: Toutes les organisations
         id: organization
-        type: organization
+        type: organization_custom
+        search_display: basic
+        values: []
```
abulte: Remove

configs/ecospheres/config.yaml
```diff
@@ -490,10 +420,21 @@ pages:
         display: true
     editable: false
     filters:
+      - name: Mise à jour
+        id: last_update_range
+        type: last_update_range
+        search_display: basic
+        values: []
```
abulte: Remove

configs/ecospheres/config.yaml
```diff
@@ -490,10 +420,21 @@ pages:
...
+      - name: Label
+        id: badge
+        type: badge
+        search_display: basic
+        values: []
```
abulte: Remove

configs/ecospheres/config.yaml
```diff
@@ -627,6 +572,98 @@ pages:
...
+        create: true
+      datasets:
+        display: true
+    editable: false
+    filters:
+      - name: Mise à jour
+        id: last_update_range
+        type: last_update_range
+        search_display: basic
+        values: []
```
abulte: Remove

configs/ecospheres/config.yaml
```diff
@@ -627,6 +572,98 @@ pages:
...
+      - name: Organisation
+        default_option: Toutes les organisations
+        id: organization
+        type: organization_custom
+        search_display: basic
+        values: []
```
abulte: Remove

configs/ecospheres/config.yaml
```diff
@@ -627,6 +572,98 @@ pages:
...
+        display: true
+      reuses:
+        display: true
+    editable: true
+    filters:
+      - name: Date de mise à jour
+        id: last_update_range
+        type: last_update_range
+        search_display: basic
+        values: []
```
abulte: Remove

cypress/e2e/custom/culture/datasets/list.cy.ts
```diff
@@ -30,92 +29,53 @@ describe('Culture - Datasets List Page', () => {
...
+      // drains one queued call before triggering the action (to skip the page-load request).
+      // That queued call was already consumed above, so a second drain would time out.
+      cy.selectFilterValue("Type d'organisation", 'Association')
+      // Verify the API call contains both filter parameters
+      cy.wait('@get_datasets_list').then((interception) => {
+        expect(interception.request.url).to.include('tag=livre')
+        expect(interception.request.url).to.include(
+          'organization_badge=association'
+        )
+      })
```
abulte: Make it a param `drain = true` in `expectActionToCallApi` instead

cypress/e2e/custom/ecospheres/datasets/list.cy.ts
```diff
@@ -70,17 +61,14 @@ describe('Datasets - List Page', () => {
...
-      .contains('Indicateur')
-      .should('be.visible')
-
-    // Check that the regular dataset does NOT have the "Indicateur" badge
-    cy.contains('Regular Dataset Title')
-      .closest('li')
-      .find('.fr-badge')
-      .should('not.exist')
+    // The indicator card wraps both the badge and the DatasetCard.
+    // Use the wrapper as the anchor to avoid fragile up-traversal across component boundaries.
```
abulte:
```suggestion
    // Check that the indicator dataset has the "Indicateur" badge
```

cypress/e2e/custom/ecospheres/topics/list.cy.ts
```diff
@@ -67,18 +58,10 @@ describe('Topics - List Page', () => {
...
-          cy.get('.multiselect-wrapper').click()
-        })
-
-      // Select "Mieux consommer" from the dropdown options
-      cy.get('.multiselect-option[aria-label="Mieux consommer"]').click()
+      cy.selectFilterValue('Thématique', 'Mieux consommer')
 
-      // Verify the URL contains the theme parameter
-      cy.url().should('include', 'theme=mieux-consommer')
+      // SearchSelectFilter stores the full prefixed value in the URL
```
abulte:
```suggestion
      // Verify the URL contains the theme parameter
```

cypress/e2e/custom/ecospheres/topics/list.cy.ts
```diff
@@ -90,7 +73,9 @@ describe('Topics - List Page', () => {
     })
   })
 
-  describe('Draft Checkbox', () => {
+  // TODO: private filter not yet ported to UnifiedSearchView.vue
+  // (was in deleted TopicList.vue — wire it up via #custom-filters-bottom slot)
```
abulte:
```suggestion
```

cypress/e2e/custom/ecospheres/topics/list_empty.cy.ts
```diff
@@ -5,42 +5,40 @@ describe('Topics - Empty List and Reset Filters', () => {
...
-      // Start with empty results
-      cy.mockDatagouvObjectList('topics', [])
       cy.visit('/bouquets')
       cy.wait('@get_topics_list')
 
       // Verify "Réinitialiser les filtres" button appears when no results
       cy.contains('button', 'Réinitialiser les filtres').should('be.visible')
 
       // Navigate away and come back with all filters applied to test the reset
+      // SearchSelectFilter stores the full prefixed value in the URL
```
abulte:
```suggestion
```

cypress/e2e/custom/ecospheres/topics/list_empty.cy.ts
```diff
@@ -5,42 +5,40 @@ describe('Topics - Empty List and Reset Filters', () => {
...
-      // Verify the theme filter is visually selected in the UI
+      // Verify the theme filter is visually selected (ComboboxInput shows the selected value)
       cy.contains('label.fr-label', 'Thématique')
-        .parent('.fr-select-group')
-        .within(() => {
-          cy.get('.multiselect-single-label').should('be.visible')
-        })
+        .closest('.fr-input-group')
+        .find('input')
+        .should('not.have.value', '')
```
abulte: This seems brittle as a test

abulte: Same below.

cypress/e2e/custom/simplifions/cas_d_usages_list.cy.js
```diff
@@ -98,15 +103,23 @@ describe("Simplifions Cas d'usages Listing Page", () => {
...
+          'simplifions-v2-types-de-simplification-acces-facile',
+          'simplifions-v2-cas-d-usages'
+        ]
+      }
     )
   })
 
-  it('should not have the private filter', () => {
+  // TODO: private filter not yet ported to UnifiedSearchView.vue
+  // (was in deleted TopicList.vue — wire it up via #custom-filters-bottom slot)
```
abulte:
```suggestion
```

cypress/e2e/custom/simplifions/solutions_list.cy.js
```diff
@@ -95,15 +103,23 @@ describe('Simplifions Solutions Page', () => {
...
+          'simplifions-v2-types-de-simplification-acces-facile',
+          'simplifions-v2-solutions'
+        ]
+      }
     )
   })
 
-  it('should not have the private filter', () => {
+  // TODO: private filter not yet ported to UnifiedSearchView.vue
+  // (was in deleted TopicList.vue — wire it up via #custom-filters-bottom slot)
```
abulte:
```suggestion
```

cypress/support/filters.js
```diff
@@ -32,10 +24,33 @@ Cypress.Commands.add('clickCheckbox', (checkbox_name) => {
...
+    }
+
+    const alias = `@get_${resourceName}_list`
+
+    // Drain one initial call (page load), then trigger the action.
+    cy.wait(alias)
     action()
-    cy.wait(`@get_${resourceName}_list`)
+
+    // Pages with multiple search types (e.g. simplifions) fire one request per
```
abulte:
```suggestion
    // Pages with multiple search types fire one request per
```

src/components/search/SearchOrganizationFilter.vue
abulte, line 1: Could this be inherited from `src/components/search/SearchSelectFilter.vue`? Not sure it's really warranted.

src/components/topics/TopicCard.vue
```diff
@@ -26,6 +26,14 @@ const props = defineProps({
...
     required: true
   },
+  topicUrl: {
+    type: Object as () => RouteLocationRaw,
+    default: null
+  },
+  organizationUrl: {
+    type: Object as () => RouteLocationRaw,
+    default: null
+  },
```
abulte: Check if it's actually used. If not, remove.

src/custom/accessibilite/routes.ts
```diff
@@ -12,13 +11,11 @@ const topicConf = {
...
 
 export const routes: RouteRecordRaw[] = [
-  useDatasetSearchPageRoutes({
-    pageKey: 'datasets',
-    metaTitle: 'Données'
+  useGlobalSearchPageRoutes({
+    pageKey: 'datasets'
   }),
-  useTopicSearchPageRoutes({
+  useGlobalSearchPageRoutes({
```
abulte, line 16: Big refactor, not only for accessibilite, let's talk about this. Could we create the globalsearchpageroutes directly from config for all pages? Is `topicConf` still used differently by different sites? Could we move the relevant topicConf stuff to config instead?

src/custom/defis/routes.ts
```diff
@@ -11,13 +10,11 @@ const topicConf = {
...
-  useDatasetSearchPageRoutes({
-    pageKey: 'datasets',
-    metaTitle: 'Données'
+  useGlobalSearchPageRoutes({
+    pageKey: 'datasets'
   }),
-  useTopicSearchPageRoutes({
+  useGlobalSearchPageRoutes({
     pageKey: 'defis',
-    metaTitle: 'Défis',
```
abulte, line 20: Any idea why we needed this in meta rather than in config before?

src/model/config.ts
```diff
@@ -32,24 +32,41 @@ export interface PageFilterFormConf {
   required: boolean
 }
 
+/** GlobalSearch-native filter type keys (passed directly to basicFilters/advancedFilters) */
+export enum GlobalSearchNativeFilterType {
```
abulte: Can't be extracted from upstream?

abulte: If not, add FIXME

src/router/utils.ts
```diff
@@ -7,6 +17,49 @@ import {
...
+  ReturnType<typeof getDefaultDataserviceConfig>['hiddenFilters']
+>[number]
+type TopicFilterKey = NonNullable<
+  ReturnType<typeof getDefaultTopicConfig>['basicFilters']
+>[number]
+type TopicHiddenFilter = NonNullable<
+  ReturnType<typeof getDefaultTopicConfig>['hiddenFilters']
+>[number]
+
+export interface CustomSelectFilterConfig {
```
abulte:
```suggestion
export interface SelectFilterConfig {
```

Plus associated refactor

src/router/utils.ts
```diff
@@ -7,6 +17,49 @@ import {
...
+  values: Array<{ value: string; label: string }>
+}
+
+export interface OrganizationFilterConfig {
+  urlParam: 'organization'
+  label: string
+  defaultLabel: string
+  apiParam: 'organization'
+  pageKey: string
+}
```
abulte: Do the hardcoded attributes `urlParam` and `apiParam` make the typing union below better? If not, why?

src/router/utils.ts
```diff
@@ -73,125 +126,30 @@ export const useRouteQueryAsString = (): RouteLocationQueryAsString => {
...
-    datasetCardComponent,
-    detailsViewComponent,
-    descriptionComponent,
-    // loosen type before sending to generic fn
-    props: topicConf as unknown as Record<string, unknown>
-  })
+  // GlobalSearch-specific
+  searchType?: PageObjectType
+  searchConfig?: GlobalSearchConfig
+  customSelectFilters?: CustomFilterConfig[]
```
abulte:
```suggestion
  customFilters?: CustomFilterConfig[]
```

Plus associated refactor

src/router/utils.ts
```diff
@@ -202,18 +160,14 @@ export const useSearchPageRoutes = ({
...
         },
         component: listViewComponent,
-        props: (route: RouteLocationNormalizedLoaded) => ({
-          // this forces the component to be recreated when switching page type
-          key: pageKey,
-          query: route.query.q,
-          page: route.query.page,
-          ...props
+        props: () => ({
+          // no props needed — UnifiedSearchView reads everything from the route directly
```
abulte: Remove

src/router/utils.ts
```diff
@@ -227,7 +181,7 @@ export const useSearchPageRoutes = ({
           datasetCardComponent
         },
         props: () => ({
-          // this forces the component to be recreated when switching page type
+          // forces component remount when switching between page types
```
abulte:
```suggestion
          // this forces the component to be recreated when switching page type
```

src/router/utils.ts
```diff
@@ -261,6 +215,190 @@ export const useTopicAdminPagesRoutes = ({
...
+  topicConf?: TopicPageRouterConf
+}
+
+const NATIVE_FILTER_TYPE_SET = new Set<string>(
+  Object.values(GlobalSearchNativeFilterType)
+)
+
+const datasetSortOptions = [
+  { value: '-created' as const, label: 'Date de création' },
+  { value: '-last_update' as const, label: 'Dernière mise à jour' }
```
abulte: `as const` needed?

src/router/utils.ts
```diff
@@ -261,6 +215,190 @@ export const useTopicAdminPagesRoutes = ({
...
+  const pageConf = usePageConf(pageKey)
+  const allPages = usePagesConf()
+  const searchConfig: GlobalSearchConfig = []
+  for (const [key, conf] of Object.entries(allPages)) {
+    if (!conf.list_all) continue
+    searchConfig.push(buildSingleTypeConfig(key, conf.object_type))
+  }
+
+  // Build customSelectFilters for the primary page (rendered via SearchSelectFilter/SearchOrganizationFilter in #custom-filters slot).
+  const customSelectFilters: CustomFilterConfig[] = (pageConf.filters ?? [])
```
abulte: Why are we parsing `pageConf.filters` here AND in `buildSingleTypeConfig`?

src/typings/route-meta.d.ts
```diff
@@ -7,12 +10,13 @@ declare module 'vue-router' {
...
     cardComponent?: () => Promise<{ default: Component }>
     datasetCardComponent?: () => Promise<{ default: Component }>
-    listComponent?: () => Promise<{ default: Component }>
     descriptionComponent?: () => Promise<{ default: Component }>
-    cardClass?: string
     pageKey?: string
+    // GlobalSearch-specific
+    searchType?: PageObjectType
+    searchConfig?: GlobalSearchConfig
+    customSelectFilters?: CustomFilterConfig[]
```
abulte: Why are we passing both searchConfig and customSelectFilters? Isn't customSelectFilters embedded in searchConfig in the end?

src/utils/component.ts
```diff
@@ -21,13 +21,6 @@ interface UseAsyncComponentOptions<T extends Component | null = null> {
  *   () => meta?.cardComponent,
  *   { fallback: DatasetCard }
  * )
- *
- * @example
- * // Load component from router meta with loading component
- * const ListComponent = useAsyncComponent(
- *   () => meta?.listComponent,
- *   { loadingComponent: ContentPlaceholder }
- * )
```
abulte, line 30: Put it back

src/utils/config.ts
```diff
@@ -2,6 +2,8 @@ import config from '@/config'
 import type { DatasetsConf, PagesConf, TopicsConf } from '@/model/config'
 import type { SiteId } from '@/model/topic'
 
+export const usePagesConf = (): PagesConf => config.pages
```
abulte, line 5: Unused?

src/views/UnifiedSearchView.vue
```diff
@@ -0,0 +1,185 @@
...
+const userStore = useUserStore()
+const labels = computed(() => useLabels(pageConf.value.labels))
+
+const links = computed(() => [
+  { to: '/', text: 'Accueil' },
+  { text: pageConf.value.breadcrumb_title ?? pageConf.value.title }
+])
+
+// route.meta is reactive — read properties directly or via computed so they
+// update when Vue Router reuses this component across pages.
```
abulte: Comment disconnected from code.

src/views/UnifiedSearchView.vue
```diff
@@ -0,0 +1,185 @@
...
+  () => route.meta.cardComponent as (() => Promise<Component>) | undefined
+)
+
+// localType is a plain ref (not prop-backed) so GlobalSearch can mutate it freely.
+// Synced to pageKey on route change so the type selector stays in sync when the
+// component is reused by Vue Router (no remount).
+const localType = ref(pageKey.value)
+watch(pageKey, (newKey) => {
+  localType.value = newKey
+})
```
abulte: Explain better

src/views/UnifiedSearchView.vue
```diff
@@ -0,0 +1,185 @@
...
+    const { page, ...restQuery } = route.query
+    await router.push({ name: newType, query: restQuery })
+    await nextTick()
+    // Both scrollBehavior (router) and this manual restore are needed:
+    // scrollBehavior prevents Vue Router from scrolling to top on navigation,
+    // but the browser still clamps scrollY when the document shrinks during
+    // the reactive re-render (old results → loading state → new results).
+    window.scrollTo(0, scrollY)
+  }
+})
```
abulte, line 74: Cascade of watch pageKey -> localType needed?

src/views/UnifiedSearchView.vue
```diff
@@ -0,0 +1,185 @@
...
+<template>
+  <div class="fr-container">
+    <DsfrBreadcrumb class="fr-mb-1v" :links="links" />
+  </div>
+
+  <div class="fr-container fr-my-2v">
+    <div class="fr-grid-row fr-grid-row--middle justify-between fr-mb-3w">
+      <h1 class="fr-mb-0">{{ pageConf.title }}</h1>
+      <div
+        v-if="userStore.canAddTopic(pageKey)"
```
abulte: Guard is ambiguous, should check we're on topic type directly.

src/views/UnifiedSearchView.vue
```diff
@@ -0,0 +1,185 @@
...
+    </div>
+  </section>
+
+  <div class="fr-container fr-mb-4w">
+    <h2 v-if="pageConf.search?.input" class="fr-mb-2w">
+      {{ pageConf.search.input }}
+    </h2>
+    <Suspense>
+      <GlobalSearch v-model:type="localType" :config="route.meta.searchConfig!">
+        <!-- FIXME: validate placement top/bottom -->
```
abulte: Remove.

src/custom/defis/routes.ts
abulte: In the old architecture each site had its own list component (e.g. `TopicList.vue`) that needed `topicConf` values to customise its rendering — `route.meta` was the standard mechanism for passing per-route config down to components.

With `UnifiedSearchView` there is no longer a per-site list component that reads these values: the single generic view gets everything it needs from YAML via `usePageConf`. So `topicConf` dropped out of `meta` entirely and now only goes as direct `props` to `TopicFormView` via `useTopicAdminPagesRoutes`, which is the right place for a static per-route value.

Same story across all 8 sites — the consolidation into `UnifiedSearchView` made the `meta` storage redundant.

src/components/search/SearchOrganizationFilter.vue
abulte: Done.

src/components/topics/TopicCard.vue
abulte: Done.

src/router/utils.ts
abulte: Removed.

src/custom/accessibilite/routes.ts
abulte: TopicConf still used differently for different sites. Don't feel like refactoring into config now (plus, still a problem to put object specific keys in generic pages). The routes api is clean and we'll always need somewhere js/ts to inject the custom components anyway.

src/utils/config.ts
abulte: Not unused — `usePagesConf` is imported and used in `buildGlobalSearchConfig` in `src/router/utils.ts` to iterate over all pages when building the search config.

src/router/utils.ts
abulte: Needed for type narrowing re GlobalSearch

src/router/utils.ts
abulte: Different purposes: `buildSingleTypeConfig` extracts native filter type keys (e.g. `last_update_range`, `badge`) to pass as `basicFilters`/`advancedFilters` directly to GlobalSearch's internal config. `buildGlobalSearchConfig` extracts the custom filter configs (`select`, `organization_custom`) to render as Vue components in the `#custom-filters-top` slot. The two passes are over the same array but produce completely different outputs for separate consumers.

src/typings/route-meta.d.ts
abulte: They're architecturally separate. `searchConfig` is consumed entirely by GlobalSearch internally — it drives native filter rendering, result fetching, and sort options. `customFilters` are our own Vue components rendered in the `#custom-filters-top` slot; they write to URL params via `useSearchFilter` and are invisible to GlobalSearch's internals. So no, `customFilters` is not embedded in `searchConfig` — they're two different extension points.

abulte:
## Changes since last update

### `router/utils` refactor

- **Removed `useSearchPageRoutes`** — inlined into `useGlobalSearchPageRoutes` (its only caller). Eliminates the `SearchPageRoutesOptions` interface and the `props: topicConf as unknown as Record<string, unknown>` cast.
- **Removed 6 filter type aliases** (`DatasetFilterKey`, `DataserviceFilterKey`, etc.) — replaced with inline `Parameters<typeof X>[0]` casts in `buildSingleTypeConfig`. The FIXME note moved to the cast site.
- **Shared `baseArgs`** in `buildSingleTypeConfig` — the three branches (dataset / dataservice / topic) no longer repeat the same 5 fields individually.
- **`defaultDetailsView` ternary** → `Record<PageObjectType, …>` map.
- **`defaultLabel` made optional** in `SelectFilterConfig` / `OrganizationFilterConfig` — fallbacks (`'Tous'`, `'Toutes les organisations'`) now live in `SearchSelectFilter.vue` and `SearchOrganizationFilter.vue` respectively, where they belong as UI defaults.
- **Dropped `?? []`** guards on `pageConf.filters` — `PageConf.filters` is already typed as `PageFilterConf[]` (non-optional).

### `SearchOrganizationFilter` refactored

Now delegates entirely to `SearchSelectFilter` (reuses its `useSearchFilter` binding, URL param handling, and `SearchableSelect` rendering) instead of duplicating them. Removes the custom `{ id, name }` model type.

### `simplifions` header search removal (completed)

- `header.display: false` in simplifions config
- `useSearchEndpoint` removed from `TopicPageRouterConf` (dead, was only used for simplifions topic search which now goes through GlobalSearch like everything else)
- Accessibilite / simplifions routes cleaned up accordingly
- Remaining dropdown test stubs removed

### `TopicCard` — `organizationUrl` prop removed

Dead prop: the router link to the organization page was removed, only the plain `OrganizationNameWithCertificate` remains.

### `UnifiedSearchView` fixes

- `canAddTopic` create button now gated on `object_type === 'topics'` (was showing for all pages)
- `customSelectFilters` renamed to `customFilters` throughout (`route-meta.d.ts`, template slot)
- `useAsyncComponent` JSDoc updated

### Tests

- `expectActionToCallApi` gains a `{ drain: false }` option for chained filter assertions (avoids draining a second initial call when already consumed)
- Culture dual-filter test rewritten using the new option (no more inline `cy.wait` + manual URL assertion)
- `list_empty.cy.ts` topic filter assertions strengthened: now checks the actual visible value (`'Mieux consommer'`, `'ADEME'`) instead of `not.have.value: ''`

## What's next

- **FIXMEs** — `GlobalSearchNativeFilterType` and the filter interfaces (`basicFilters` / `hiddenFilters` arg shapes) are inferred from upstream return shapes rather than imported. Both FIXMEs note that `@datagouv/components-next` needs to export these directly.

DorineLam (dismissed):
Ok pour simplifions

narduin (dismissed):
huge work well done!

TuThoThai (dismissed):
approved for PAUDA based on review made by @narduin

configs/ecospheres/config.yaml
```diff
@@ -234,7 +233,8 @@ definitions: &definitions
 #       plural: {string}            # plural name for the object
 #       extended: {string}          # extended name for the object, singular
 #     search:
-#       input: {string}             # placeholder for the search input
+#       input: {string}             # label displayed above the search input
+#       placeholder: {string|null}  # input placeholder (null = no placeholder, omit = upstream default)
```
streino:
```suggestion
#       placeholder: {string?|null?}  # input placeholder (null = no placeholder, omit = upstream default)
```
Also, clarify what's "upstream" here?

eudespeyre (dismissed):

doc/datagouv-components-dev.md
streino, line 1: ✨

package.json
```diff
@@ -33,7 +33,7 @@
     "type-check": "vue-tsc --noEmit --project tsconfig.app.json"
   },
   "dependencies": {
-    "@datagouv/components-next": "^1.0.1",
+    "@datagouv/components-next": "1.0.2-dev.67",
```
streino: Release 1.0.2 before merging this PR.

configs/ecospheres/config.yaml
```diff
@@ -292,14 +293,28 @@ pages:
         display: true
     editable: true
     filters:
+      - name: Mise à jour
+        id: last_update_range
+        type: last_update_range
+        search_display: basic
+      - name: Label
```
streino:
```suggestion
      - name: Label de donnée
```
Same as datagouv

configs/ecospheres/config.yaml
```diff
@@ -292,14 +293,28 @@ pages:
         display: true
     editable: true
     filters:
+      - name: Mise à jour
```
streino:
```suggestion
      - name: Date de mise à jour
```
Same as datagouv

configs/ecospheres/config.yaml
```diff
@@ -292,14 +293,28 @@ pages:
...
+        type: badge
+        search_display: basic
       - name: Organisation
         default_option: Toutes les organisations
-        id: organization
-        type: organization
+        id: org
+        type: organization_custom
+        search_display: basic
+      - name: Granularité
```
streino:
```suggestion
      - name: Granularité spatiale
```
Same as datagouv

configs/ecospheres/config.yaml
```diff
@@ -490,6 +416,14 @@ pages:
         display: true
     editable: false
     filters:
+      - name: Mise à jour
```
streino:
```suggestion
      - name: Date de mise à jour
```

configs/ecospheres/config.yaml
```diff
@@ -627,6 +564,95 @@ pages:
...
+        [Restez informés des dernières publications via notre lettre d'information](https://grist.numerique.gouv.fr/o/ecolabservicesdonnees/forms/mCtZDvP6vKbMcorvXM7sGK/55)
+    resources_tabs:
+      discussions:
+        display: true
+        create: true
+      datasets:
+        display: true
+    editable: false
+    filters:
+      - name: Mise à jour
```
streino:
```suggestion
      - name: Date de mise à jour
```

src/components/CheckboxComponent.vue
streino, line 1: Are we removing this because it won't work anymore, or because it's not used with the removal of brouillon?

cypress/e2e/custom/ecospheres/datasets/list.cy.ts
```diff
@@ -35,23 +34,61 @@ describe('Datasets - List Page', () => {
...
+      'datasets',
+      { tag: 'adresses' }
+    )
+
+    cy.get('@countBefore').then((countBefore) => {
+      cy.get('@get_dataservices_list.all').should('have.length', countBefore)
+    })
+  })
+
+  it('should clear org filter when switching to a type with a different org list', () => {
```
streino, line 78: Are we sure about that? If the org is still in the new list, I would expect that org to stay selected. That's what datagouv does between dataset and API.

src/model/config.ts
```diff
@@ -32,24 +32,44 @@ export interface PageFilterFormConf {
...
-    | 'spatial_zone'
-    | 'spatial_granularity'
     | 'select'
-    | 'checkbox'
-    | 'organization'
-    | 'private'
+    // `organization_custom` in order to differentiate from native upstream filter
+    | 'organization_custom'
+    | GlobalSearchNativeFilterType
+  /** Where to display this filter in GlobalSearch (undefined = not shown in GlobalSearch) */
```
streino: Nitpick: `/** ... */` format inconsistent for this kind of comments. 

Only other occurrences (that are not function doc headers) are in cypress/e2e/custom/ecospheres/topics/support.ts.

src/views/UnifiedSearchView.vue
```diff
@@ -0,0 +1,191 @@
...
+    </h2>
+    <Suspense>
+      <GlobalSearch v-model:type="localType" :config="route.meta.searchConfig!">
+        <template v-if="route.meta.customFilters?.length" #custom-filters-top>
+          <template
+            v-for="filter in route.meta.customFilters"
+            :key="filter.label"
+          >
+            <SearchSelectFilter v-if="'values' in filter" :config="filter" />
+            <SearchOrganizationFilter v-else :config="filter" />
```
streino: Feels brittle to fallback to `SearchOrganisationFilter`. I don't know how feasible in templates, but I'd instead do ifs for different types of filters, and fallback to an error. That'll make it obvious something's wrong if we add another type of filter without `'values'`.

Edit: A more defensive way to handle fallback would definitely make sense, esp. [fallback isn't always falling on the same case](https://github.com/opendatateam/udata-front-kit/pull/1153/changes#r3209607661). IMO would also read better: no need to remember/lookup there are only two custom types.

src/router/utils.ts
```diff
@@ -70,197 +99,230 @@ export const useRouteQueryAsString = (): RouteLocationQueryAsString => {
   return { ...route, query }
 }
 
-interface SearchPageRoutesOptions {
+export const useTopicAdminPagesRoutes = ({
```
streino: Not sure why this got moved. Could stay near the end of file and avoid the diff.

src/router/utils.ts
```diff
@@ -70,197 +99,230 @@ export const useRouteQueryAsString = (): RouteLocationQueryAsString => {
...
+): GlobalSearchConfig[number] {
+  const pageConf = usePageConf(pageKey)
+  const hiddenFilters = Object.entries(pageConf.universe_query ?? {}).map(
+    ([key, value]) => ({ key, value })
+  )
+  const basicFilters: string[] = []
+  const advancedFilters: string[] = []
+  for (const filter of pageConf.filters) {
+    if (!NATIVE_FILTER_TYPE_SET.has(filter.type)) continue
+    if (!filter.search_display) {
```
streino: I'd go with `advanced` true/false and default to false. No risk of forgetting and simpler to reason about. Also see https://github.com/opendatateam/udata-front-kit/pull/1153#discussion_r3209708776.

src/router/utils.ts
```diff
@@ -70,197 +99,230 @@ export const useRouteQueryAsString = (): RouteLocationQueryAsString => {
...
+    )
+  }
+  if (searchType === 'dataservices') {
+    return getDefaultDataserviceConfig(
+      baseArgs as Parameters<typeof getDefaultDataserviceConfig>[0]
+    )
+  }
+  return getDefaultDatasetConfig({
+    ...baseArgs,
+    sortOptions: datasetSortOptions
```
streino: Why only datasets? Is it not configurable for other types (weird), or do other types have acceptable defaults (might change, so let's force them like we do for datasets), or ...?

configs/ecospheres/config.yaml
```diff
@@ -490,6 +416,14 @@ pages:
         display: true
     editable: false
     filters:
+      - name: Mise à jour
+        id: last_update_range
+        type: last_update_range
+        search_display: basic
+      - name: Label
```
streino:
```suggestion
      - name: Label de donnée
```

src/router/utils.ts
```diff
@@ -70,197 +99,230 @@ export const useRouteQueryAsString = (): RouteLocationQueryAsString => {
...
+          f.type === 'organization_custom')
+      if (isRenderable && !f.search_display) {
+        console.warn(
+          `Filter "${f.id}" on page "${pageKey}" has no search_display — it will not be shown. Add search_display: basic or advanced to the config.`
+        )
+      }
+      return isRenderable && !!f.search_display
+    })
+    .map((f) => {
+      if (f.type === 'organization_custom') {
```
streino, line 199: Also same as https://github.com/opendatateam/udata-front-kit/pull/1153/changes#r3209395954, except this is the inverse case (if org else values), so a new unknown type will fallback the other way vs the other places where test occurs.

src/router/utils.ts
```diff
@@ -70,197 +99,230 @@ export const useRouteQueryAsString = (): RouteLocationQueryAsString => {
...
-    listComponent,
-    filtersComponent,
-    cardComponent,
-    detailsViewComponent
-  })
-}
+  // Build customFilters for the primary page (rendered via SearchSelectFilter/SearchOrganizationFilter in #custom-filters slot).
+  const customFilters: CustomFilterConfig[] = pageConf.filters
+    .filter((f) => {
+      const isRenderable =
```
streino: I'm not sure about the term "renderable". This suggests other types are not renderable? Except when `hide_on_list` is true, what's the point of a non-renderable filter? Sent me down the wrong path before I figured it's only a way to filter non-hidden custom filters.

Becomes moot with https://github.com/opendatateam/udata-front-kit/pull/1153/changes#r3209465426: you wouldn't need to test `search_display` and therefore wouldn't need the intermediate variable at all.

src/router/utils.ts
```diff
@@ -70,197 +99,230 @@ export const useRouteQueryAsString = (): RouteLocationQueryAsString => {
...
+        !f.hide_on_list &&
+        ((f.type === 'select' && f.values?.length) ||
+          f.type === 'organization_custom')
```
streino: Same as https://github.com/opendatateam/udata-front-kit/pull/1153/changes#r3209395954

src/router/utils.ts
```diff
@@ -7,6 +17,25 @@ import {
...
+  typeKeys: string[]
+}
+
+export interface OrganizationFilterConfig {
+  label: string
+  defaultLabel?: string
+  pageKey: string
+  urlParam: string
+  typeKeys: string[]
+}
```
streino: `SelectFilterConfig` and `OrganizationFilterConfig` have a few common parameters. Might want to extract a common super, or at least order params in the same way, so similarities are more obvious.

streino (commented):
Reviewed code only, didn't (yet) test. PR looked a lot scarier that it really was, congrats on making the code so clean to refactor :)

configs/ecospheres/config.yaml
```diff
@@ -490,6 +416,14 @@ pages:
...
+        id: badge
+        type: badge
+        search_display: basic
```
streino: Do we want to keep it?

src/router/utils.ts
streino: It might also be worth standardizing options over all types to preserve sort order when switching from one type to the other. Currently it's preserved when possible (e.g. -last_update on dataset -> indicateur) but dropped otherwise (dataset -> API), so if you're not paying attention it's a bit unsettling.

cypress/e2e/custom/ecospheres/datasets/list.cy.ts
streino: Definitely feels weird:
1. Go to datasets and select org=ADEME
2. Type updates and shows 7 API
3. Click on type=API
4. Filter drops, API is updated to 86

Same on datagouv will keep filter and show only the 7 API on step 4.

configs/ecospheres/config.yaml
streino: Something fishy here? Config says "Label" but deployed preview shows "Label de donnée" as captured in https://github.com/opendatateam/udata-front-kit/pull/1153/changes#r3209831721.

streino (commented):
Here come the tests. Overall works really well!

configs/ecospheres/config.yaml
abulte: Yeah, doesn't make sense here.

configs/ecospheres/config.yaml
abulte: This one goes away, as discussed with @Thesauruv.

src/components/CheckboxComponent.vue
abulte: Both!

cypress/e2e/custom/ecospheres/datasets/list.cy.ts
abulte: It's easier for data.gouv.fr, they have the same lists of orgs for every type. We don't. So the type switch and background search would need to be extra-smart: does the currently selected value exist in the other type? and its behaviour would be different depending on which org is selected. Plus, the boundary is unclear: this should belong to upstream but it doesn't. I'm not sure the complexity is warranted at this point, maybe keep this as a future feature if it's really a problem.

src/router/utils.ts
abulte: Per the specs, datasets was the only one that needed to change. We can't use the same options for all types, since dataservice doesn't support modified date sort. That leaves Topics, but not sure it's worth enforcing anything here. I moved the sort options inline, might be less disturbing.

configs/ecospheres/config.yaml
abulte: Yep, name is unused :-/

abulte: Native filters don't support a custom label upstream. I think we can leave with that for now (since we were trying to align with data.gouv.fr's wording anyway). I added a FIXME/doc in config.

package.json
abulte: I will ask for a release once it's stabilised here (making sure no more upstream changes needed).

configs/ecospheres/config.yaml
```diff
@@ -293,28 +296,20 @@ pages:
...
-        search_display: basic
-      - name: Granularité
-        id: granularity
-        type: granularity
-        search_display: advanced
       - name: Thème INSPIRE
         default_option: Tous les thèmes
         id: inspire
         type: select
-        search_display: advanced
```
streino: Not `advanced: true`?

src/model/config.ts
```diff
@@ -1,3 +1,8 @@
+import type { BuiltInFilterKey } from '@datagouv/components-next'
+
+export const CUSTOM_FILTER_TYPES = ['select', 'organization_custom'] as const
+export type CustomFilterType = (typeof CUSTOM_FILTER_TYPES)[number]
```
streino, line 4: Feels convoluted to have to go through `CUSTOM_FILTER_TYPES` as we only should need `CustomFilterType`? (might be a TS thing, but weird)

src/router/utils.ts
```diff
@@ -223,44 +191,49 @@ export function buildGlobalSearchConfig(pageKey: string): {
...
-      if (isRenderable && !f.search_display) {
-        console.warn(
-          `Filter "${f.id}" on page "${pageKey}" has no search_display — it will not be shown. Add search_display: basic or advanced to the config.`
-        )
-      }
-      return isRenderable && !!f.search_display
+      if (f.hide_on_list) return false
+      if (!CUSTOM_FILTER_TYPE_SET.has(f.type as CustomFilterType)) return false
+      return true
     })
```
streino: For some reason I can't do suggestions anymore... 

Could simplify to `(f) => !f.hide_on_list && CUSTOM_FILTER_TYPE_SET.has(f.type as CustomFilterType)`

src/router/utils.ts
```diff
@@ -223,44 +191,49 @@ export function buildGlobalSearchConfig(pageKey: string): {
...
-          pageKey,
-          urlParam: f.id,
-          typeKeys: [pageKey]
+        return [
+          {
+            label: f.name,
+            defaultLabel: f.default_option ?? undefined,
+            pageKey,
+            urlParam: f.id,
+            typeKeys: [pageKey]
```
streino: Nitpick: I'd use the same param order as defined in the type for readability, and/or at least the same order with 'select' below.

configs/ecospheres/config.yaml
streino: We can live with that, and I actually think it's a good thing! So I wouldn't consider this FIXME. 

Could we make `name` optional (only required for custom filters) and have `id` as the key (makes more sense anyway)? That way we don't even have to set `name` for native filters.

src/router/utils.ts
streino: It does look less disturbing in the code this way. Behavior will still be weird, but that's not something we can fix unless we use the same sort options everywhere. Datagouv doesn't so I guess at least we're not inconsistent with them.

cypress/e2e/custom/ecospheres/datasets/list.cy.ts
streino: I agree it's trickier and would be better handled upstream; though it might be difficult to push it upstream since they don't have the issue.

Maybe we should revisit how we generate the orgs lists? It made sense to have separate lists to hide 0-match orgs when types were clearly separate. But with unified search it might make more sense to have a single (union'ed) list, with some types simply showing 0 match.

Either way this is out of scope here, but might be worth discussing with @Thesauruv (when you're back).

configs/ecospheres/config.yaml
```diff
@@ -222,50 +221,54 @@ definitions: &definitions
...
+#       plural: {string}              # plural name for the object
+#       extended: {string}            # extended name for the object, singular
 #     search:
-#       input: {string}             # placeholder for the search input
-#     banner:                       # banner for the list page (optionnal)
-#       title: {string}             # title for the banner
-#       content: {string}           # content for the banner
-#     resources_tabs:               # resources tabs at the bottom of a detail page
+#       input: {string}               # label displayed above the search input
+#       placeholder: {string?|null?}  # input placeholder (null = no placeholder, omit = data.gouv.fr's default)
```
streino:
```suggestion
#       placeholder: {string?|null?}  # input placeholder (null = no placeholder, omit = data.gouv.fr default)
```

well, suggestions are back...

configs/ecospheres/config.yaml
```diff
@@ -222,50 +221,54 @@ definitions: &definitions
...
-#           - id: {string}          # id of the value ⚠️ needs to be unique among all values of the filter, across all parent values
-#             name: {string}        # name of the value
-#             parent: {string?}     # id of the parent filter value (optional)
+#         display: {bool}             # display reuses tab on detail page
+#     editable: {bool}                # whether the page is editable when connected
+#     filters:                        # list of filters for this {page_id}
+#       - name: {string}              # name of the filter  ⚠️ needs to be unique across all items
+#                                     # FIXME: the name is not used for data.gouv.fr native filters (eg Label de donnée)
+#         id: {string}                # id of the filter
+#         type: {string}              # type of the filter (select | organization_custom | or any GlobalSearchNativeFilterType)
```
streino: Refer to `CustomFilterType` instead of enumerating them here? Could however be useful to have pointers to where it's defined (applies also for native) so we know what we can use.

configs/ecospheres/config.yaml
abulte: Nope, this was noop, works only for native filters.

src/model/config.ts
abulte: Yeah, can't find a better way without duplicating the values. We need CUSTOM_FILTER_TYPES elsewhere to iterate on the values.

configs/ecospheres/config.yaml
abulte: I'll keep the FIXME because I'm pretty sure we'll have some change requests on that, and I prefer consistent keys as much as possible (even if noop for now).

src/model/config.ts
streino: Right, TS doesn't do runtime types, so makes sense...

configs/ecospheres/config.yaml
streino: Worth adding to the doc

configs/ecospheres/config.yaml
streino: Makes sense to keep it FIXME in current state of things. How much work would it be to switch to id keys though?

configs/ecospheres/config.yaml
abulte: Already is!

configs/ecospheres/config.yaml
```diff
@@ -627,6 +551,90 @@ pages:
...
+    filter_prefix: ecospheres
+    title: Collections thématiques
+    breadcrumb_title: Collections
+    labels:
+      singular: collection
+      plural: collections thématiques
+      extended: collection thématique
+      feminine: true
+    search:
+      input: Filtrer les collections thématiques
```
streino:
```suggestion
      input: Rechercher une collection thématique
```
Uniformiser avec les autres types qui sont tous en "rechercher xxx"

configs/ecospheres/config.yaml
abulte: There's no key in the config format as such, but you mean smtg like that? https://github.com/opendatateam/udata-front-kit/pull/1153/commits/f0531735659e766764d23df27b310871b5c7705f

src/views/UnifiedSearchView.vue
streino, line 1: Upstream issue but mentioning it here in case we can do something about it (or push it upstream). Might also affect custom filter behavior.

When selecting a value for a filter that doesn't apply to all types, only the types relevant to the filter are updated, and the other types remain with the same previous count. IMO the count for non-relevant types should drop to 0 (or N/A).

For instance, if you select a specific "Modalité d'accès" on datagouv, reuses and organizations keep their count. Same on ecologie for "Levier" in indicateurs. It looks even worse if you pick a "levier" with no matches, because then indicateurs count drops to 0, but all other counts are still pretending to have matches:

configs/ecospheres/config.yaml
streino: Uh? I meant mentioning the flag applies only to native filters.

```
advanced: {bool?}           # show in the advanced (collapsed) section of GlobalSearch; omit for basic (always visible)
```

configs/ecospheres/config.yaml
streino: Yes! I'm using "key" a bit too loosely: I meant something that needs to be unique to be discriminatory. I'm surprised there's only [a single location](https://github.com/opendatateam/udata-front-kit/commit/f0531735659e766764d23df27b310871b5c7705f#diff-1264fef71025b0b0a40069be6fbe0ef7653cd0e989277ad1a91b1dd5e03acaaaR136) where this requires a change, nice!

So `name` can become optional then? I really think we should drop those entries when they are noop to avoid confusion. We still can keep the FIXME (or just a note) in the doc to raise awareness on the "issue" for other verticals.

streino (commented):
Sorry, still finding some rough edges...

configs/ecospheres/config.yaml
abulte:
```
#         advanced: {bool?}           # show in the advanced (collapsed) section of GlobalSearch; omit for basic (always visible)
#                                     # ⚠️ only applicable to data.gouv.fr native filters (eg badge)
```

src/views/UnifiedSearchView.vue
abulte: That's a tricky one. I agree the current behaviour might feel strange, but I'm not sure your solution is really better. I feel this is a limitation of the unified UI with different filters per type.

At least current behaviour is coherent: every type only ever fires a query or computation with its own filters, and the common ones are shared (except organization which is a tricky one for us, as we've seen elsewhere). I guess it might also help the user figure out / intuit that a filter only applies to a given type.

I'll leave it like that for now. This warrants a UX research on data.gouv.fr's side.

configs/ecospheres/config.yaml
abulte: Like I said, I like having the config keys as homogenous as possible across filters. I also like having the name (even if it could diverge, but there's a FIXME) to search for the filter. I'll leave it like that, `id` is now first-class in the config order.

src/views/UnifiedSearchView.vue
streino: Yes I didn't expect changes here, but we should bring it up with datagouv. I agree 0 isn't ideal. I think we still would need a way to signal the scope of a change and what is out of it. N/A, graying out, ...

configs/ecospheres/config.yaml
streino: I disagree about having "pretend" properties that make you think is doing something but isn't. This IMO is less intuitive than having an optional property with a clear logic (native filter: no property / custom: needs property).

But the key point here is that I think we really shouldn't consider this FIXME (my previous comment wasn't clear about that). It's *good* that we enforce the datagouv label, and we shouldn't suggest we'll change that. If there's a strong push to have it configurable we'll reconsider, but we should not plan for it, and IMO should resist against it, esp. since we've had so much feedback about things being different from datagouv.

So I'd prefer if `name` was optional but won't argue further about it, but I'd really like the FIXME to be changed to a note or something that doesn't convey the impression that we'll add a way to customize the lable.

configs/ecospheres/config.yaml
streino: :raised_eyebrow: Why it didn't show in the diff I have no idea... I see it now :ok_hand:

configs/ecospheres/config.yaml
streino: Could use the same as https://github.com/opendatateam/udata-front-kit/pull/1153#discussion_r3224233417 instead of the FIXME (still keeping it required if you prefer it so):
```
name: {string}      # name of the filter
                    # ⚠️ only applicable to custom filters
```

package.json
abulte: Done.

cypress/e2e/custom/ecospheres/datasets/list.cy.ts
abulte: Out of scope here indeed, let's discuss this later.

configs/ecospheres/config.yaml
abulte: I don't really want to but let's meet halfway :p Done.

streino (approved):
woot! :D

PR #1168, opened 2026-05-04
fix(ecologie): hp search reset

### LINKED ISSUE #1025: Query vide après une recherche à partir de la page dd'accueil
Reproduction : 
- Je formule une requête dans la barre de la page d'accueil
- J'appuie sur "Enter" sur le clavier (sans cliquer directement sur le bouton rechercher)
- J'arrive dans l'onglet toutes les données, mais la requête n'est pas conservée.

https://github.com/user-attachments/assets/cdae3457-f9ef-4171-a871-8c46f09c364c

abulte:
Fix https://github.com/ecolabdata/ecospheres/issues/1025

Due to double submit in DsfrSearchBar when using Enter: one when pressing Enter, one for form submit. No FIXME because our code stays clean.

### DIFF
--- a/src/custom/ecospheres/components/home/HomeHero.vue
+++ b/src/custom/ecospheres/components/home/HomeHero.vue
@@ -10,8 +10,9 @@ const router = useRouter()
 const searchQuery = ref('')
 
 const doSearch = (q: string) => {
-  router.push({ name: 'datasets', query: { q } })
-  searchQuery.value = ''
+  router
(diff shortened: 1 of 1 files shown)


streino (approved):
Might be worth a comment in the code?

abulte:
> Might be worth a comment in the code?

Wondered that myself, but the code is clean as is, so I'll just count on the PR text 🤞

PR #1171, opened 2026-05-04
feat(indicators): vendor-in viz component

### LINKED ISSUE #1093: Visualisation indicateurs : intégrer à udata-front-kit

abulte:
Fix https://github.com/ecolabdata/ecospheres/issues/1093

This migrates the logic of https://github.com/ecolabdata/ecospheres-tabular-dataviz (will need to be archived) into the front-kit.

- iterating on bug and features will be way faster in a single repo
- the idea of a shared lib with other projects is long gone
- the migration lifts the code quality up to this repo standards

This is not a full rewrite, the core logic has been kept (and probably some bugs too). But it uses vuejs paradigms instead of DOM manipulation and Typescript instead of JS. The contract/interface is also simplified, and we have more tests. I plan further refactors when future features or fixes are implemented.

I hope there aren't any regression, but it's hard to test all the cases. I did review all the closed bugs relating to viz and tested the fixes still apply.


src/custom/ecospheres/views/indicators/IndicatorDetailView.vue
streino, line 1: I have a couple update issues on https://deploy-preview-1171--ecospheres.sandbox.data.developpement-durable.gouv.fr/indicators/67d7ff439ae15029b56e4336

1. From the default state (Région, Regrouper):
   1. Switch to EPCI => ❌ Only legend shows on a blank page with even Regrouper gone (sometimes it does work, in that case keep switching Maille).
   2. Select another Territoire value => ✔️ Number properly shows.

2. From the default state (Région, Regrouper): 
   1. Disable Regrouper => ✔️ Number switches to graph.
   2. Switch do Département => ✔️ Graph updates.
   3. Switch to EPCI => ✔️ Graph updates.
   4. Switch back to Département => ❌ Only legend shows on blank page.

3. From the default state (Région, Regrouper):
   1. Disable Regrouper => ✔️ Number switches to graph.
   2. Switch to EPCI => ❌ Regrouper is back to enabled.

I tried the same tests several times and didn't alway get the same behavior, but eventually end up with a Regrouper inconsistency or with a "blank page with only legend".

streino (changes requested):

src/custom/ecospheres/views/indicators/IndicatorDetailView.vue
streino: Firefox linux in case it matters

abulte: Thanks for the thorough tests! Fortunately, they were all linked and fixed by better handling the territory change (race conditions basically).

src/custom/ecospheres/components/indicators/viz/useIndicatorVizChart.ts
```diff
@@ -0,0 +1,160 @@
...
+  watch(series, (newSeries) => createOrUpdateChart(newSeries), {
+    flush: 'post'
+  })
+  watch(canvasRef, (canvas) => {
+    if (!canvas) destroyChart()
+  })
+
+  onBeforeUnmount(destroyChart)
+
+  return { destroyChart }
```
streino: Return not used in only call to that function. Why/when would it be needed?

streino (dismissed):

src/custom/ecospheres/components/indicators/viz/useIndicatorVizChart.ts
abulte: Could be useful if lifecycle were a bit more complicated, but we can add it back if needed.

streino (approved):

PR #1173, opened 2026-05-06
fix(indicators): no data handling in viz

### LINKED ISSUE #1040: Prévisualisation sur une région avec des données par défaut
En lien avec cette issue https://github.com/ecolabdata/ecospheres/issues/1039#issue-4281327141

La prévisualisation est mise par défaut sur la région Auvergne Rhône Alpes. Or dans le cas où il n'y a pas de données pour ce territoire l'affichage apparaît comme cassé

Image

Fonctionnement attendu:
- Lorsqu'il n'y a pas de données afficher un message "L'indicateur n'a pas de données pour le territoire sélectionné"
- Par défaut sélectionner un territoire qui a des données pour que la prévisualisation apparaisse correctement
- Idéallement n'avoir dans la liste déroulante que des territoires présents dans le fichier

abulte:
Fix https://github.com/ecolabdata/ecospheres/issues/1040

Handles the case where no data is available for a given territory.

Lightweight solution to https://github.com/ecolabdata/ecospheres/issues/1040, I didn't do:
- auto-switch to filled-in territory
- filtering out territories with no data

It maybe feels a bit over-engineered (after all, we only need data for Auvergne Rhone Alpes for the feature to become quite useless), plus it might be disturbing for the user to not find its territory in the list without explanation. Also, would be not trivial to implement with interesting edge cases (an empty file would fire a query for all territories for example). WDYT @lcolumelli @Thesauruv?

Capture d’écran 2026-05-06 à 08 20 43

image

Example: https://deploy-preview-1173--ecospheres.sandbox.data.developpement-durable.gouv.fr/indicators/69e0e67708ce9429e6c4b551

### DIFF
--- a/cypress/e2e/custom/ecospheres/indicators/viz.cy.ts
+++ b/cypress/e2e/custom/ecospheres/indicators/viz.cy.ts
@@ -1,0 +1,79 @@
+import { createIndicator, createIndicatorResource } from './support'
+
+describe('Indicator Viz', () => {
+  beforeEach(() => {
+    cy.mockMatomo()
+    cy.mockStaticDatagouv()
+  })
+
+  describe('No Datavisualisation Configured', () => {
+    it('should not display the previsualisation', () => {
+      const indicator = createIndicator({}, { enable_visualization: false })
+      cy.mockDatasetAndRelatedObjects(indicator)
+      cy.visit(`/indicators/${indicator.id}`)
+      cy.contains('Prévisualisation').should('not.exist')
+    })
+  })
+
+  describe('Prévisualisation Tab', () => {
+    it('should display an error when the tabular API fails', () => {
+      const indicatorWithViz = createIndicator(
+        {},
+        { enable_visualization: true }
+      )
+      const vizResource = createIndicatorResource()
+      cy.mockDatasetAndRelatedObjects(indicatorWithViz, [vizResource])
+      cy.intercept(
+        'GET',
+        `https://tabular-api*.data.gouv.fr/api/resources/${vizResource.id}/data/**`,
+        { statusCode: 500, body: 'Internal Server Error' }
+      )
+      cy.visit(`/indicators/${indicatorWithViz.id}`)
+      cy.contains('Prévisualisation').click()
+      cy.contains('Erreur lors du chargement').should('be.visible')
+      cy.contains('k: millier, M: million, Md: milliard').should('not.exist')
+    })
+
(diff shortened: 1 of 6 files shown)


cypress/e2e/custom/ecospheres/indicators/viz.cy.ts
```diff
@@ -0,0 +1,79 @@
...
+      )
+      const vizResource = createIndicatorResource()
+      cy.mockDatasetAndRelatedObjects(indicatorWithViz, [vizResource])
+      cy.intercept(
+        'GET',
+        `https://tabular-api*.data.gouv.fr/api/resources/${vizResource.id}/data/**`,
+        {
+          statusCode: 200,
+          body: {
+            data: []
```
streino: Maybe add some fake data to differentiate from the previous no-data test?

src/custom/ecospheres/components/indicators/viz/IndicatorVizChart.vue
streino, line 1: You get "no data" when deselecting all categories (obviously). Might want to have a different info message for that case: 


This can also happen when you have selected only specific categories and switch to another Territoire that doesn't have the same categories. In a way the info message is true, but we should be suggesting to change the selected categories. To reproduce: 
1. Département + Alpes-de-Haute-Provence + only "Travaux, matériaux, maintenance":
   
2. Switch to "Ain":
   
3. Switch back to Alpes-de-Haute-Provence 
   => Same as step 1. 
   ⚠️ This shows an axis display issue rather than just the info message. If we preserve the selected categories when switching, they should show up even if there's no data for it.

cypress/e2e/custom/ecospheres/indicators/viz.cy.ts
abulte: We're testing `{ enable_visualization: true }` so only one of those tests is enough, merged into the "display no data" case.

src/custom/ecospheres/components/indicators/viz/IndicatorVizChart.vue
abulte: - selected axis values are now reset when territory changes (same as legacy version)
- new message when no axis is selected

Capture d’écran 2026-05-13 à 15 09 26

streino: Works well now!

Might want to sort the axis values, otherwise they are shuffled around when changing Territoire.

streino (approved):

PR #1198, opened 2026-05-15
feat: accessibilite prod config

narduin:
Hide themes page from nav and sitemap.
We can still access the url through the `/datasets` "Thèmes" filter but I don't know how to remove it or if it is necessary.

@TuThoThai is it necessary to prevent access to the themes' page?


TuThoThai:
@narduin, hiding it is enough as we hope to make it apparent soon. 😄

TuThoThai (dismissed):

abulte (dismissed):
FYI you can use `list_all: false` on the `themes` page to make it non-addressable.

abulte (approved):

TuThoThai (approved):

PR #1207, opened 2026-05-20
feat(seo): add meta description support to static pages

### LINKED ISSUE #1108: SEO : meta page statique
Ref https://docs.numerique.gouv.fr/docs/a9510bf2-ee4b-4e16-bb8a-37d6b260858b/

URL (page A propos) https://ecologie.data.gouv.fr/about 

TITLE (🔴 AVANT) : À propos | [ecologie.data.gouv.fr](https://ecologie.data.gouv.fr/) 
TITLE (🟢 APRÈS) : Notre mission | [ecologie.data.gouv.fr](https://ecologie.data.gouv.fr/)

DESCRIPTION (🔴 AVANT) : Accédez aux données de référence pour la transition écologique. Découvrez, réutilisez et partagez votre expertise pour agir en faveur d’un avenir durable grâce aux jeux de données, indicateurs et collections thématiques des services de l’État et des collectivités. DESCRIPTION (🟢 APRÈS) : [ecologie.data.gouv.fr](https://ecologie.data.gouv.fr/) met à disposition des acteurs de la planification écologique (État, collectivités, entreprises, associations, citoyens) des jeux de données actualisées, des indicateurs et des réalisations concrètes. L’objectif est de faciliter l'adaptation au changement climatique et la territorialisation des objectifs de transition écologique.

## Implementation

Frontmatter avec une description meta injectable.

abulte:
Fix https://github.com/ecolabdata/ecospheres/issues/1108

### DIFF
--- a/src/router/index.ts
+++ b/src/router/index.ts
@@ -64,7 +64,7 @@ const pages = (config.website.router.static_pages ?? []).map(
       props: { url: item.url },
       meta: {
         title: item.title,
-        metaDescription: item.meta_description
+        metaDescription: item.meta?.description
       }
     }
   }
--- a/src/router/index.ts
+++ b/src/router/index.ts
@@ -63,7 +63,8 @@ const pages = (config.website.router.static_pages ?? []).map(
       component: StaticPageView,
       props: { url: item.url },
       meta: {
-        title: item.title
+        title: item.title,
+        metaDescription: item.meta_description
       }
     }
   }
--- a/configs/ecospheres/config.yaml
+++ b/configs/ecospheres/config.yaml
@@ -130,10 +130,11 @@ website:
       - title: 'À propos'
         id: 'about'
         route: '/about'
-        meta_description: |
-          ecologie.data.gouv.fr met à disposition des acteurs de la planification écologique (État, collectivités, entreprises, associations, citoyens) 
-          des jeux de données actualisées, des indicateurs et des réalisations concrètes. L’objectif est de faciliter l'adaptation au changement 
-          climatique et la territorialisation des objectifs de transition écologique.
+        meta:
+          description: |
+            ecologie.data.gouv.fr met à disposition des acteurs de la planification écologique (État, collectivités, entreprises, associations, citoyens) 
+            des jeux de données actualisées, des indicateurs et des réalisations concrètes. L’objectif est de faciliter l'adaptation au changement 
+            climatique et la territorialisation des objectifs de transition écologique.
         url: '/static/ecospheres/pages/about.md'
       - title: 'Accessibilité'
         id: 'accessibility'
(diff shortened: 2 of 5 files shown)


src/router/index.ts
```diff
@@ -63,7 +63,8 @@ const pages = (config.website.router.static_pages ?? []).map(
       component: StaticPageView,
       props: { url: item.url },
       meta: {
-        title: item.title
+        title: item.title,
+        metaDescription: item.meta?.description
```
streino, line 67: Why not just `description`? We're in `meta` already.

abulte: Unfortunate namespace conflict, we're in `meta` but of `route.meta` (not related to HTML meta).

streino: And `title` here is both the route and html meta title?

IIUC route meta is just a bunch of stuff associated with the route, so would it be problematic to store `description` there? Esp. if I'm correct with `title` serving both?

If it's bad form to conflate the two metas (Claude says it's common, but he's probably just being nice), could we maybe go with a `htmlMeta` dict straight away and put `description` in there? Esp. if we might have other html-meta stuff later on?

streino (dismissed):

src/router/index.ts
abulte: Which meta are you complaining about? 😅 One of them I can't change (route.meta). The one in the config file is changeable but I'd rather not, it's already being used and is nice and short. This leave the meta in `metaDescription`?

streino: Well, *that* meta! ;)

I meant either my original suggestion (conflates route-meta with html-meta):
```
  meta: {  // route.meta
    title: item.title,  // IIUC used by both route logic and html meta.
    description: item.meta?.description  // only used in html meta, but why not?
  }
```

Or (subdict dedicated to html meta):
```
  meta: {
    title: item.title,  // either use it for both from here, or duplicate in html.title
    html: {  // or htmlMeta
      title: item.title,  // optional
      description: item.meta?.description,
      // whatever else we may need later? (keywords maybe?)
    }
  }
```

From what I understand, `route.meta` is just an arbitrary dict? So we can put whatever we want in there?

abulte: Alright. So `route.meta` can hold anything indeed, but I'd rather keep it simple (plain flat keys for now) — plus, html is not used in config. `metaDescription` is nice because it maps directly to the config path (`meta.description`). I don't mind `description` in that case, but it could become confusing if we use `description` for something else someday (route meta typing is shared across routes). I still have a slight preference for `metaDescription`.

streino: Let's stick with `metaDescription` then, I'll probably be the only one wondering about that lone property having a prefix :p

streino (approved):

PR #1211, opened 2026-05-25
feat(orgs): custom routes

### LINKED ISSUE #1117: Organisations list optional and from config

abulte:
Fix https://github.com/ecolabdata/ecospheres/issues/1117

`useOrganizationsRoutes` in `src/custom/ecospheres/routes.ts` instead of default routes that need to be disabled for all sites except ecologie.

`config.router.disable[]` is now useless (`topics` is dead code, artifact from pre pages migration).

While we're at, organization detail meta title is now more precise (Contributeur) and config structure is refactored a bit.

Doc is done in https://github.com/opendatateam/udata-front-kit/pull/1212

### DIFF
--- a/configs/ecospheres/config.yaml
+++ b/configs/ecospheres/config.yaml
@@ -199,7 +199,7 @@ organizations:
   datasets: https://raw.githubusercontent.com/opendatateam/udata-front-kit-universe/refs/heads/artifacts/dist/ecospheres-demo/organizations-datasets.json
   dataservices: https://raw.githubusercontent.com/opendatateam/udata-front-kit-universe/refs/heads/artifacts/dist/ecospheres-demo/organizations-dataservices.json
   bouquets: https://raw.githubusercontent.com/opendatateam/udata-front-kit-universe/refs/heads/artifacts/dist/ecospheres-demo/organizations-bouquets.json
-  list:
+  page: # organizations.page is a special case of config.pages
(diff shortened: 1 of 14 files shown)


configs/ecospheres/config.yaml
```diff
@@ -201,8 +199,10 @@ organizations:
   datasets: https://raw.githubusercontent.com/opendatateam/udata-front-kit-universe/refs/heads/artifacts/dist/ecospheres-demo/organizations-datasets.json
   dataservices: https://raw.githubusercontent.com/opendatateam/udata-front-kit-universe/refs/heads/artifacts/dist/ecospheres-demo/organizations-dataservices.json
   bouquets: https://raw.githubusercontent.com/opendatateam/udata-front-kit-universe/refs/heads/artifacts/dist/ecospheres-demo/organizations-bouquets.json
-  list:
+  page: # organizations.page is a special case of config.pages
```
abulte, line 202: I don't think it's worth moving that to `pages`, it's a _very_ special case. But I think it's nice to have a bit of coherence (config keys below exist in `pages`).

streino: 👍 
Hopefully we can move it there when we get a real org filter 🤞

streino (dismissed):

narduin (dismissed):

PR #1212, opened 2026-05-25
feat(docs): update docs for unified search

### LINKED ISSUE #1114: Mettre à jour la doc de dev re UnifiedSearch

abulte:
Fix https://github.com/ecolabdata/ecospheres/issues/1114
Depends https://github.com/opendatateam/udata-front-kit/pull/1211

For simplicity's sake, also includes the doc from https://github.com/opendatateam/udata-front-kit/pull/1211.


doc/architecture.md
```diff
@@ -130,24 +135,34 @@ filters:
...
-| `hide_on_list`   | Masque le filtre sur la page de liste (tout en l'appliquant)                     |
-| `parent`         | (sur une valeur) Identifiant de la valeur parente pour les filtres hiérarchiques |
+| Clé                | Description                                                                                                |
+| ------------------ | ---------------------------------------------------------------------------------------------------------- |
+| `default_option`   | Option sélectionnée par défaut                                                                             |
+| `default_value`    | Valeur par défaut                                                                                          |
+| `advanced`         | Place le filtre dans la section "Filtres avancés" de `GlobalSearch`                                        |
+| `applies_to_pages` | Indique les pages pour lesquelles le filtre est partagé (valeur de filtre conservée au changement de page) |
+| `hide_on_list`     | Masque le filtre sur la page de liste (tout en l'appliquant)                                               |
```
streino: Is the filter shown somewhere else than on page lists? If not, how about renaming it simply to `hidden`?

Also, it might be useful to add a small section defining "page de liste" and "page de détail" (= the basic elements other than static pages) right at the beginning of "Les univers et leur paramétrage"? You sort of build an understanding the further down you read the doc, but it'd be easier with a definition.

abulte: > Is the filter shown somewhere else than on page lists? If not, how about renaming it simply to hidden?

Yes, it's used by indicators for tags. Description is misleading, it's not applying the filter at all.

abulte: > Also, it might be useful to add a small section defining "page de liste" and "page de détail" (= the basic elements other than static pages) right at the beginning of "Les univers et leur paramétrage"? You sort of build an understanding the further down you read the doc, but it'd be easier with a definition.

I doubt it's really useful (did I get it right?) but added in https://github.com/opendatateam/udata-front-kit/pull/1212/commits/6bfa88d365dfd601471f4b712ea2e4b5ae889836

streino: > I doubt it's really useful (did I get it right?)

You did, and at the very least it can't hurt :) Thanks!

> Yes, it's used by indicators for tags. 

`hidden` would still work in this case, no? I mean such a filter doesn't show *anywhere*. 

This actually points to a semantics ambiguity around the term "filter", or rather how we bend filters to retreive indicator badges.

We call "filter" both a) the filter applied on the datagouv API, and b) the UI filter on the list page. In most cases they go together, but the "trick" for indicator badges decorrelates the two meanings when we don't want to filter on those badges.

Am I understanding this correctly?

I don't think it's worth splitting badges and filters since it really custom extra-based stuff, but using plain "hidden" instead of "hide on some specific page" would at least make it clearer that this stuff won't show in the UI?

abulte: I'm not sure `hidden` is better. It does show in the UI, as a tag; so technically, it's hidden on list. I agree it's not great (because it's not a filter), but like we already discussed, this is an anecdotal feature, not worth refactoring into something complicated. And out of scope anyway, this is just doc :-)

streino (approved):

PR #1213, opened 2026-05-25
fix(search|a11y): avoid input focus

### LINKED ISSUE #1115: UnifiedSearch input steals the focus
Focus should be in LiveRegion shortcuts on page load, not the search input.

abulte:
Depends https://github.com/datagouv/cdata/pull/1096
Fix https://github.com/ecolabdata/ecospheres/issues/1115


package.json
```diff
@@ -29,7 +29,7 @@
     "type-check:cypress": "tsc --noEmit --project cypress/tsconfig.json"
   },
   "dependencies": {
-    "@datagouv/components-next": "1.1.1",
+    "@datagouv/components-next": "1.0.2-dev.90",
```
abulte: Release stable before merging.

abulte: Semver seems broken :-/

src/views/topics/TopicFormView.vue
```diff
@@ -182,9 +182,13 @@ const onSubmit = async () => {
...
     if (props.isCreate && routeQuery.clone != null) {
-      return `Cloner ${labels.articles.le} ${labels.singular} ${topic.value.name}`
+      return topic.value.name
+        ? `Cloner ${labels.articles.le} ${labels.singular} ${topic.value.name}`
+        : undefined
     } else if (!props.isCreate) {
-      return `Éditer ${labels.articles.le} ${labels.singular} ${topic.value.name}`
+      return topic.value.name
+        ? `Éditer ${labels.articles.le} ${labels.singular} ${topic.value.name}`
+        : undefined
```
abulte, line 191: Unrelated, this was announcing / displaying undefined on mount.

streino (dismissed):
OK as far as I can tell (not much). I'd rather rely on @narduin's review.

narduin (commented):
So the page title doesn't change anymore when you do a search is that right?
We have a special title that's sent to be announced in the live region instead?

abulte:
> So the page title doesn't change anymore when you do a search is that right? We have a special title that's sent to be announced in the live region instead?

Before #1153, we had a live region announcement of the search results count (which you implemented ;-)). Since it was broken by the new component, it's reintroduced here. It's not a title, cf the debug logs in console:

```
[LiveRegion] Announcement [{"text":"Page Toutes les données pour \"xxtestxx\" | ecologie.data.gouv.fr"},{"text":"13 résultats"}]
```

narduin:
> Before #1153, we had a live region announcement of the search results count (which you implemented ;-)). Since it was broken by the new component, it's reintroduced here. It's not a title, cf the debug logs in console:

Yep I remember :grin: 
Although I thought the search results were also added to the page title, has it been too long for my memory already? :fearful:

abulte:
> > Before #1153, we had a live region announcement of the search results count (which you implemented ;-)). Since it was broken by the new component, it's reintroduced here. It's not a title, cf the debug logs in console:
> 
> Yep I remember 😁 Although I thought the search results were also added to the page title, has it been too long for my memory already? 😨

I don't remember either 😬 Should they? Seems a bit redundant if live region is updated, no?

narduin:
I think the title and the live region were both updated on route change but it should be tested with a screen reader to be sure.
If that's not the case anymore we might not need to update the live region here but could use the actual text "XX résultats" as a live region.

abulte:
> I think the title and the live region were both updated on route change but it should be tested with a screen reader to be sure. If that's not the case anymore we might not need to update the live region here but could use the actual text "XX résultats" as a live region.

"XX résultats" is now upstream, so this might be hard.

narduin:
> "XX résultats" is now upstream, so this might be hard.

It might be enough because it has a `role="status"` on it! Which is annouced when updated automatically.

```js
<p
  class="fr-col-auto my-0"
  role="status"
  data-testid="search-result-count"
>
  {{ t("{count} résultats | {count} résultat | {count} résultats", searchResults.total) }}
</p>
```

When testing this branch with a screen reader, the "live title" is annouced twice and the results are also announced since `role="status"` is an implicit live region.

```log
Page Jeux de données pour "occitqnie" | accessibilite.data.gouv.fr 2 résultats    
Page Jeux de données pour "occitqnie" | accessibilite.data.gouv.fr 2 résultats   
2 résultats
```

abulte:
@narduin so I should remove results announcement feature from this PR?

narduin:
> @narduin so I should remove results announcement feature from this PR?

I believe so, yes.

narduin (dismissed):

streino (dismissed):

PR #1216, opened 2026-05-27
feat(infra): add CSP

### LINKED ISSUE #1121: Add a Content Security Policy (CSP)
Cf https://github.com/ecolabdata/ecospheres/issues/1119

abulte:
Fix https://github.com/ecolabdata/ecospheres/issues/1121

Add a basic CSP and add a few security headers to Dockerfile (should be used by data.gouv.fr infra too). 

The main point is to protect `script-src` (which is probably the most critical). `connect-src` and `img-src` have to be open due to the nature of our content and features (cf comments below).

`style-src` now forbids the usage of inline styles 💅, which warranted some fixes in templates.

The Dockerfile is used by dokku, so the deploy previews act as test envs.

### DIFF
--- a/src/custom/simplifions/views/ApisFranceConnectees.vue
+++ b/src/custom/simplifions/views/ApisFranceConnectees.vue
@@ -4,161 +4,258 @@
 
     <div class="fr-grid-row fr-grid-row--gutters">
       <div class="fr-col-12 fr-col-lg-8">
-      <p class="fr-text--lead">Les API FranceConnectées donnent accès à diverses données administratives des particuliers en proposant FranceConnect comme modalité d'appel.</p>
-      <p class="fr-mb-4w fr-text--lg">Elles permettent de simplifier les démarches d'un particulier utilisant FranceConnect en récupérant automatiquement d'autres informations administratives le concernant.
-      </p>
+        <p class="fr-text--lead">
+          Les API FranceConnectées donnent accès à diverses données
+          administratives des particuliers en proposant FranceConnect comme
+          modalité d'appel.
+        </p>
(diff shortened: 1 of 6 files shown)


abulte:
Re `img-src`: it's very wide and kind of has to be. It should even be totally open if we consider that users often include images in their datasets' descriptions by linking to an external hosting (~~still the case in 2026?~~ yep, confirmed*). It's also hard to maintain, we have to list at least all the relevant OVH datacenters for example.

The most problematic domains are:
- raw.githubusercontent.com
- gist.github.com
- \*.s3.\*.io.cloud.ovh.net

because anyone can create a ressource on there.

That said, `img-src` is mostly used to XSS-exfiltrate data through GET parameters. I don't think GitHub lets you see the access logs and exploit the data, so we might be alright there (pending a vulnerability on GitHub or smtg). OVH is more problematic, I think users can access their buckets logs. So maybe we should tighten it to known bucket types X known envs X know datacenters (minimum 1 x 2 x 3 = 6).

Or just `*`.

➡️ went with `*` since it's needed for md content and fighting the origin domains is not worth it.

\* Capture d’écran 2026-05-27 à 17 53 40

abulte:
Same for `connect-src`: at least for the WMS preview, we need to able to connect to arbitrary domains.

src/custom/simplifions/views/ApisFranceConnectees.vue
abulte, line 1: Diff is bigger than it should be because the linter/formatter never passed on this file apparently.

streino (approved):
Rubber stamping to allow merge, but relying on @jordanguedj for the actual review 🙏

PR #1220, opened 2026-05-29
feat: accessibilite subpages

narduin:
Add about and glossary pages and their styles.

fix https://github.com/opendatateam/PAUDA/issues/5
fix https://github.com/opendatateam/PAUDA/issues/7

### DIFF
--- a/src/assets/main.css
+++ b/src/assets/main.css
@@ -382,13 +382,34 @@ body {
     font-weight: normal;
     text-transform: uppercase;
   }
+  dd + dt {
+    margin-block-start: 1rem;
+  }
 }
 
 /* ===== Lists ===== */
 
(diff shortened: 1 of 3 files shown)


src/assets/main.css
```diff
@@ -382,15 +382,36 @@ body {
...
+/* Remove list styles on ul, ol elements with a list role, which tells browsers the list still behaves as a list even without default styles */
 :where(ol, ul)[role='list'] {
   list-style: none;
 }
 
+dl > dt {
+  font-weight: bold;
+}
+
+.accessibilite-glossary {
```
abulte:
```suggestion
.glossary {
```

Makes more sense in a non-specific style file and could be reused I guess.

abulte (approved):

PR #1221, opened 2026-05-29
feat: header conf refactor

narduin:
Add header config.
@abulte our design includes both `serviceLogo` and `serviceTitle` but they are exclusive right now.
I'll try to make it possible without breaking everyone else's config :grimacing:

This might need to be a normal PR to `main` and not a sub PR to accessibilite only :innocent: 

## Refactor how header and footer logos are handled

- use `footer` and `header` keys to handle their logos
- change `service_logo` to `{header, footer}.logo` to be a secondary logo if needed (like in meteo)
- change `operator_logo` to `title_image` to better reflect its use
- allow displaying the service title along the header logo (like in accessibilite)
- fallback to `website.title` if no `title_image` is set
- use native beta badge
- migrated all configs to match the new logic

> [!WARNING]
>  Need to merge https://github.com/opendatateam/udata-front-kit/pull/1219 first then rebase here.


abulte (commented):
I'd like to propose a refactor 😇 I find it way too complicated to reason about right now.

What do you think of the following conf:

```
  header:
    logo: { src, width, height }    # brand icon, optional
    title:
      src: '...'    # wordmark image; if absent → website.title text
      width: '...'
      height: '...'
      description: '...'    # subtitle, optional
    beta: true|false    # use dsfr component beta instead of custom badge

  footer:
    logo: { src, width, height }    # optional, independent
```

This would remove:
- website.service_logo
- website.logo_operator
- website.formatted_title (dead, never read)
- website.badge
- website.header.show_title
- website.header.service_description

And we don't have to worry about what _operator_ and _service_ really mean.

narduin:
> ```yml
>   header:
>     logo: { src, width, height }    # brand icon, optional
>     title:
>       src: '...'    # wordmark image; if absent → website.title text
>       width: '...'
>       height: '...'
>     description: '...'    # subtitle, optional
>     beta: true|false    # use dsfr component beta instead of custom badge
> 
>   footer:
>     logo: { src, width, height }    # optional, independent
> ```

Yes this is great, I just have a doubt on the naming of `title: src: '…'` as it might not be expected to be an image with the key "title"

abulte:
> Yes this is great, I just have a doubt on the naming of `title: src: '…'` as it might not be expected to be an image with the key "title"

True. I find it acceptable because it ties the config value to the result (the image _is_ a title).

narduin:
> True. I find it acceptable because it ties the config value to the result (the image _is_ a title).

How about `title_image` then ? :grin: 
It makes it clear when used in templates as well.

abulte:
> > True. I find it acceptable because it ties the config value to the result (the image _is_ a title).
> 
> How about `title_image` then ? 😁 It makes it clear when used in templates as well.

But we already have title as parent key? This would be `title.title_image`, which I find a bit redondant. `title.image_src` maybe?

narduin:
> But we already have title as parent key? This would be `title.title_image`, which I find a bit redondant. `title.image_src` maybe?

I meant :

```yml
header:
  title_image:
    src: '...'    # wordmark image; if absent → website.title text
    width: '...'
    height: '...'
```

abulte:
> > But we already have title as parent key? This would be `title.title_image`, which I find a bit redondant. `title.image_src` maybe?
> 
> I meant :
> 
> ```yaml
> header:
>   title_image:
>     src: '...'    # wordmark image; if absent → website.title text
>     width: '...'
>     height: '...'
> ```

Oh! But then, the fact that `description` is in there is now a problem, since it's used for the text version. And I find it less clear that it's driving the title block. But then again, the title text is not defined in there anyway... I have a slight preference for `title.image_src` I think but if you want to go with `title_image`, I don't mind.

narduin:
> Oh! But then, the fact that `description` is in there is now a problem, since it's used for the text version. And I find it less clear that it's driving the title block. But then again, the title text is not defined in there anyway... I have a slight preference for `title.image_src` I think but if you want to go with `title_image`, I don't mind.

I would put `description` outside of `title` regardless of its name because we can have a description (baseline in DSFR) without a title, no?

abulte:
> > Oh! But then, the fact that `description` is in there is now a problem, since it's used for the text version. And I find it less clear that it's driving the title block. But then again, the title text is not defined in there anyway... I have a slight preference for `title.image_src` I think but if you want to go with `title_image`, I don't mind.
> 
> I would put `description` outside of `title` regardless of its name because we can have a description (baseline in DSFR) without a title, no?

That would strange in practice, since the baseline is supposed to be displayed under the text title I think. We'd have `brand_logo | text_logo | baseline`? But why not, if it makes sense, even if nobody uses it this way.

And anyway I was wrong, description is not under title. Is `baseline` a better conf key btw?

configs/accessibilite/config.yaml
```diff
@@ -29,48 +29,52 @@ website:
...
-              url: 'themes?theme=accessibilite-des-lieux'
-              image_url: '/static/accessibilite/assets/accessibilite-des-lieux.jpg'
-            - name: 'Numérique'
-              description: "Les données en lien avec l'accessibilité numérique"
-              url: 'themes?theme=accessibilite-numerique'
-              image_url: '/static/accessibilite/assets/accessibilite-numerique.jpg'
         sub_section_tiles:
         sub_section_buttons:
     meta_description: accessibilite.data.gouv.fr référence et centralise les données relatives à l'accesssibilité.
+  footer:
```
abulte, line 44: I'd put `footer` after `header` in config file.

src/components/header/HeaderComponent.vue
```diff
@@ -36,54 +37,46 @@ const closeModal = () => {
...
-const showLogoOperatorInHeader =
-  config.website.logo_operator?.show_in_header ?? true
-const showBadge = config.website.badge.display
-const badgeText = config.website.badge.text
-const badgeStyle = config.website.badge.style
+const serviceDescription = config.website.header.description ?? undefined
+const titleImage = config.website.header.title_image?.src
+const titleImageHeight = config.website.header.title_image?.height
+const titleImageWidth = config.website.header.title_image?.width
+const showBeta = config.website.header.beta
```
abulte: Could be nice to use a `useHeaderConf()` from `src/utils/config.ts` with a model in `src/utils/config.ts` (and inject into `WebsiteConfig` that already types `header.search`).

The other vars above could benefit from `WebsiteConfig` typing.

Or just pull a complete `useWebsiteConfig(): WebsiteConfig` with header and footer in it, up to you.

src/components/header/HeaderComponent.vue
```diff
@@ -105,12 +98,21 @@ const badgeStyle = config.website.badge.style
 </template>
 
 <style scoped>
```
abulte, line 91: I think we need some styling for the beta badge to stay `blue-cumulus` after our migration (vs `green-emeraude` from vuedsfr).

src/App.vue
```diff
@@ -87,15 +88,17 @@ onMounted(() => {
...
-const footerMandatoryLinks = config.website.footer_mandatory_links
+const headerLogo = config.website.header.logo?.src
+const headerLogoHeight = config.website.header?.logo?.height
+const headerLogoWidth = config.website.header?.logo?.width
+const footerLogo = config.website.footer.logo?.src
+const footerLogoHeight = config.website.footer.logo?.height
+const footerLogoWidth = config.website.footer.logo?.width
+const footerPhrase = config.website.footer.phrase
+const footerExternalLinks = config.website.footer.external_links
+const footerMandatoryLinks = config.website.footer.mandatory_links
```
abulte: Same here, let's use a `useFooterConfig()` or `useWebsiteConfig()` and type the config.

abulte (commented):
Thanks, already clearer and cleaner 🙏

configs/accessibilite/config.yaml
narduin: I wanted to as well but after `header` there often is the related conf `menu` then `router` and `sitemap`.
I felt they should stay together, but then the footer would be a bit far from `header`, so I placed it above :sweat_smile: 
What do you think?

abulte: Ah yes, right. No good solution here, why not footer above after all. IRL we're searching for keys anyway.

src/model/config.ts
```diff
@@ -125,13 +126,34 @@ export type DatasetsConf = {
...
+  description?: string
+  beta: boolean
+  search: {
+    display: boolean
+    placeholder?: string
+  }
+}
+
+export type FooterConf = {
+  logo: Logo
```
abulte:
```suggestion
  logo?: Logo
```

src/model/config.ts
```diff
@@ -125,13 +126,34 @@ export type DatasetsConf = {
...
+interface Logo {
+  src?: string
+  width?: string
+  height?: string
 }
 
-export interface WebsiteConfig {
+export type HeaderConf = {
+  admin_shortcut: boolean
+  logo: Logo
```
abulte:
```suggestion
  logo?: Logo
```

src/model/config.ts
```diff
@@ -125,13 +126,34 @@ export type DatasetsConf = {
...
+  title_image: Logo
```
abulte:
```suggestion
  title_image?: Logo
```

abulte (commented):
Just a few remarks re types and I think we're good.

abulte (dismissed):
Thanks! Can you maybe change the PR title to mention the refactor?

narduin:
> Thanks! Can you maybe change the PR title to mention the refactor?

yes good idea, done!
I had also forgot to remove `webp` images that got put back from the rebase on `main` 🫥

abulte (approved):

TuThoThai (approved):
Thanks for this work, @narduin! Love the refactor!

PR #1238, opened 2026-06-05
chore: upgrade deps

abulte:
Process:

1. `pnpm audit --fix update`
2. audit entries of `minimumReleaseAgeExclude` cf https://github.com/pnpm/pnpm/issues/11563: all unnecessary 

(`pnpm audit --fix update` runs `pnpm update` under the hood, so some patch updates are persisted in package.json)

3. Cypress was upgraded manually (fixes some transitive vulns)
4. unhead/vue was upgraded manually (v1 branch is unmaintained)
5. xmlbuilder2 was upgraded manually (fixes transitive vulns) -> this introduces a stub-hack re `events` lib (required by lib but not available to browser); we should drop the xmlbuilder2 dep at some point. We still upgrade it here to avoid relying on a vulnerable package.


abulte:
AI analysis re 1.:

```
  Nothing should break at runtime. Detailed findings:

  - maplibre-gl 5.3→5.13: v5.11.0 has a "breaking" label but it's an internal API rename
  (SourceCache→TileManager) — the release notes explicitly say it's not a public API break.
  The project only uses Map, Marker, and StyleSpecification, all unaffected.
  - vueuse 12.4→12.8: v12.8.1 was flagged as breaking, but it's actually the inverse — it
  reverts an accidental breaking change from 12.8.0 (useEventSource readonly). v12.8.0 also
  deprecates MaybeRef/MaybeRefOrGetter in favor of Vue's native types, but deprecation only —
  no removal.
  - @sentry/vue 10.17→10.27: No breaking changes in any of the 10 versions between them.
  - @gouvfr/dsfr, vue, pinia, marked, axios: All patch/small minor bumps with no breaking
  changes.

  One thing worth a quick check: eslint 9.18→9.39 and typescript-eslint 8.20→8.48 are large
  dev dep jumps that could introduce new lint rules and fail CI even though they won't break
  the app. Worth running pnpm run lint before pushing.
  ```

**Comment by abulte** [comment]
Only one vuln should remain from the 8 on https://github.com/opendatateam/udata-front-kit/security/dependabot:

```
  Summary: this branch now closes 7 of 8 alerts. The only one remaining is #75 minimatch
  [HIGH] — pinned by @vue/test-utils → js-beautify → editorconfig@1.x. That's upstream and
  can't be fixed without an override hack.
```

streino (approved):

PR #1241, opened 2026-06-08
feat(indicators): refactor tabs

### LINKED ISSUE #1139: Refactoring onglet ressources des indicateur
Image

Les métadonnées secteurs et leviers sont déplacées sous le titre de l'indicateur.
Les métadonnées enjeux et thématiques sont totalement retirée des fiches et tuiles indicateurs.
Métadonnée maille : inchangée par rapport au comportement actuel.

abulte:
Fix https://github.com/ecolabdata/ecospheres/issues/1139
Related https://github.com/ecolabdata/ecospheres/issues/915

I also added `next_expected_update_quarter` support in informations tab, it provides a nice balance with frequency. We'll see in https://github.com/ecolabdata/ecospheres/issues/915 if moving/copying it to the sidebar is still warranted.

NB: the tags accordion is done with minimal effort (reuse extras even though the structure is different). YMMV, I don't think it matters a lot (power user feature).

Capture d’écran 2026-06-08 à 12 01 49
Capture d’écran 2026-06-08 à 12 17 44

### DIFF
--- a/src/custom/ecospheres/components/indicators/IndicatorInformationPanel.vue
+++ b/src/custom/ecospheres/components/indicators/IndicatorInformationPanel.vue
@@ -1,10 +1,14 @@
 <script setup lang="ts">
 import { formatDate } from '@/utils'
-import { DateRangeDetails } from '@datagouv/components-next'
+import {
+  CopyButton,
+  DateRangeDetails,
+  ExtraAccordion,
+  LicenseBadge
+} from '@datagouv/components-next'
 import { computed, toRef } from 'vue'
 import type { Indicator } from '../../model/indicator'
 import { UNFILLED_LABEL, useIndicatorExtras } from '../../utils/indicator'
-import IndicatorTags from './IndicatorTags.vue'
 import InformationPanelItem from './informations/InformationPanelItem.vue'
 import InformationPanelSection from './informations/InformationPanelSection.vue'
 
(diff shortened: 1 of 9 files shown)


streino:
CC @Thesauruv. Who should we add people from the Indicators team for such reviews?

src/custom/ecospheres/components/indicators/IndicatorInformationPanel.vue
```diff
@@ -24,47 +28,87 @@ const { unite, mailles } = useIndicatorExtras(indicator)
...
+          indicator.extras[
+            'ecospheres-indicateurs'
+          ].next_expected_update_quarter.replace('Q', 'T')
+        "
       />
     </InformationPanelSection>
+
+    <!-- Tags -->
+    <div class="pt-6">
+      <ExtraAccordion
```
streino, line 107: Having a list here doesn't make much sense: numbering has no meaning. IMO we should keep the previous "bag of tags" display. Only now it's nested in an accordion.

src/custom/ecospheres/components/indicators/IndicatorInformationPanel.vue
```diff
@@ -24,47 +28,87 @@ const { unite, mailles } = useIndicatorExtras(indicator)
...
+        v-if="indicator.temporal_coverage"
+        title="Couverture temporelle"
+      >
+        <DateRangeDetails :range="indicator.temporal_coverage" />
       </InformationPanelItem>
     </InformationPanelSection>
 
-    <!-- Métadonnées -->
-    <InformationPanelSection title="Métadonnées">
+    <!-- Détails techniques -->
```
streino, line 47: I'd group all the temporal info together, so maybe just move ID/license up or down?

src/custom/ecospheres/components/indicators/IndicatorInformationPanel.vue
streino, line 1: With the tab merge in this PR I wonder if we should go even further and merge "Sources" into "Informations" as well? This would result in a single metadata tab, which IMO would be even nicer. @Thesauruv?

src/custom/ecospheres/components/indicators/IndicatorInformationPanel.vue
abulte: Yeah but like I said, it's made to be as simple as possible. We reuse the extras accordion which expects an object. Alternative is `{tags: []}` but reads worse.

> NB: the tags accordion is done with minimal effort (reuse extras even though the structure is different). YMMV, I don't think it matters a lot (power user feature).

Bag of tags might be nice, but it means introducing a custom accordion component.

src/custom/ecospheres/components/indicators/IndicatorInformationPanel.vue
abulte: @Thesauruv

src/custom/ecospheres/components/indicators/IndicatorInformationPanel.vue
streino: Ah right... IMO we'd be better off just keeping keywords as they were, if they really have to be in an accordion, then so be it :shrug:

abulte: Can put them back as they were, ie not in accordion, without too much effort. Up to you @Thesauruv.

src/custom/ecospheres/components/indicators/IndicatorInformationPanel.vue
Thesauruv: I agree with @streino, I would move ID/license down.

abulte: Done.

src/custom/ecospheres/components/indicators/IndicatorInformationPanel.vue
Thesauruv: Agree with @streino -> keeping them as they were, though at the bottom of the page brefore the extras.

src/custom/ecospheres/components/indicators/IndicatorInformationPanel.vue
Thesauruv: Sources were initially displayed below the ressources, though users didn't see them and kept asking about them. Moving them in a dedicated tab improved their visibility. I would let as it is now.

src/custom/ecospheres/components/indicators/IndicatorInformationPanel.vue
abulte: So in their own dedicated subsection (which subsection title and which label for the value)? Or as a new value in the Détails techniques section?

abulte: Went with 2nd solution.

Capture d’écran 2026-06-11 à 18 06 27

streino (approved):

PR #1249, opened 2026-06-12
feat: accessibilite preprod fixes

narduin:
# To address comments on #1239 

- Add summary option to static pages
- hide themes page
- add new themes icon
- fix email contact address
- add custom styles through a class on `<main>`


TuThoThai (dismissed):
LGTM, @narduin !

src/utils/index.ts
```diff
@@ -31,6 +35,29 @@ export const fromMarkdown = (value: string | null, inline: boolean = false) => {
...
+  marked.use(gfmHeadingId(), {
+    hooks: {
+      postprocess(html) {
+        headings = getHeadingList()
+        return html
+      }
+    }
+  })
+  const fn = inline ? marked.parseInline : marked.parse
+  const parsed = fn(value, markedOptions)
```
abulte, line 42: Did you try using `fromMarkdown()` directly here? Seems like `marked.use` should apply. This would avoid repeating the parsing.

src/views/StaticPageView.vue
```diff
@@ -30,6 +31,22 @@ const links = computed(() => [
   { text: pageConfig.value?.title || '' }
 ])
 
+const showSummary = computed(() => pageConfig.value?.summary?.display)
+const isInlineSummary = computed(() => pageConfig.value?.summary?.inline)
```
abulte, line 35: Do we need this `inline` config? Seems like it's only used as inline.

src/views/StaticPageView.vue
```diff
@@ -30,6 +31,22 @@ const links = computed(() => [
...
+const titleSummary = computed(() => pageConfig.value?.summary?.title)
+
+const headingList = computed(() => {
+  if (!showSummary.value) return
+
+  const { headings } = fromMarkdownWithHeadings(content.value)
+  // Only keep h2s because DsfrSummary does not support nesting
+  const h2s = headings?.filter((hx) => hx.level === 2)
```
abulte: Nitpick: could be upstreamed to `fromMarkdownWithHeadings` as an optional `maxTitleLevel` param.

src/views/StaticPageView.vue
```diff
@@ -46,8 +63,27 @@ watchEffect(async () => {
...
+  </div>
   <GenericContainer>
-    <!-- eslint-disable-next-line vue/no-v-html -->
-    <div class="editorial" v-html="fromMarkdown(content)" />
+    <!-- eslint-disable vue/no-v-html -->
+    <div
+      class="editorial"
+      v-html="
+        showSummary
+          ? fromMarkdownWithHeadings(content).html
```
abulte: Parsing is done twice, once for heading, once for content.

src/utils/index.ts
abulte: Given https://github.com/opendatateam/udata-front-kit/pull/1249/changes#r3411550069, we could even make this title-extraction feature an option of `fromMarkdown`. This would lead to refactoring the existing call sites of fromMarkdown to handle an object return value, but I think it's manageable.

narduin: Yes, I initially included it in `fromMarkdown` but then saw that it was used **everywhere** by **everyone** and did not want to change that much in an accessibilite PR 😬 
Should I do it anyway ?

src/views/StaticPageView.vue
narduin: This is an option to make the summary links inline like on [the glossary page.](https://deploy-preview-1249--accessibilite.sandbox.data.developpement-durable.gouv.fr/glossaire)

src/views/StaticPageView.vue
narduin: It could but the component will not render it correctly as in the DSFR specs.
Should we allow it nonetheless ?

src/views/StaticPageView.vue
abulte: Did not see that, ok!

src/views/StaticPageView.vue
abulte: I mean upstream the filtering in `fromMarkdownWithHeadings`, nothing related to DSFR?

src/utils/index.ts
abulte: Yeah, if it only changes the signature I think it's fine.

src/utils/index.ts
```diff
@@ -20,28 +20,14 @@ const markedOptions = {
  */
 export const descriptionFromMarkdown = (ref: Ref, attr = 'description') => {
   if (ref.value?.description) {
```
abulte: Unrelated, but this should check `[attr]` exists, not `description`.

JeSuisUnCaillou (dismissed):
Ok for simplifions I guess :smile:

abulte (approved):

src/views/StaticPageView.vue
narduin: I left it there because vue-dsfr's summary component can't render nested titles properly.
If someone wants to use the headings for something else I'd rather let them have the full list available.

PR #1277, opened 2026-06-25
feat(meteo): add conditional custom filter for clim data

maudetes:
Following https://github.com/datagouv/api-meteo/pull/3, we are using Tabular API.
Custom filters fail due to huge number of rows (currently limited to 50000 on Tabular) for MIN and HOR datasets.

Seeing the very low usage rate of these custom filters, we deactivate it entirely.


Pierlou (approved):
Perfect 🤝

abulte (approved):

PR #1287, opened 2026-06-26
Meteo france preprod patch merge

maudetes:


maudetes:
See [this PR](https://github.com/opendatateam/udata-front-kit/pull/1286) instead

PR #1296, opened 2026-07-02
fix(dataset): proper harvest/non-harvest dates

streino:
Display different set of dates depending harvest status.

**Non harvested :** Display datagouv dates.

**Harvested :** Display harvest dates. Exemple has no harvested creation date => do not display "Création".

### DIFF
--- a/src/components/datasets/DatasetTemporalitySection.vue
+++ b/src/components/datasets/DatasetTemporalitySection.vue
@@ -17,31 +17,32 @@ const props = defineProps<{
 const { formatDate } = useFormatDate()
 
 const harvest = props.dataset.harvest as TypedHarvest
-const harvestCreatedAt = harvest?.created_at
-const harvestIssuedAt = harvest?.issued_at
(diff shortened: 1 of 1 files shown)


abulte (commented):
Isn't it a problem than non-harvested datasets never have creation date with this change? eg https://ecologie.data.gouv.fr/datasets/5530fbacc751df5ff937dddb.

Maybe we need a conditional on harvest for this logic.

streino:
> Isn't it a problem than non-harvested datasets never have creation date with this change? eg https://ecologie.data.gouv.fr/datasets/5530fbacc751df5ff937dddb.
> 
> Maybe we need a conditional on harvest for this logic.

Very good point. I think it'll just make more sense to have separate logic for harvested and non-harvested. Will do.

streino:
Done. Updated PR title and description. Keeping branch name because github would force me to create a new PR otherwise.

abulte (approved):

PR #1327, opened 2026-07-31
fix(ci): check formatting

abulte:
Adds formatting check as a mandatory CI step. This is because some devs/tooling would bypass the precommit locally and another correctly configured environment would reformat the files and create discrepancies between branches (eg when deploying, deploy script on a correctly configured machine would reformat the deploy branch).

The deploy script and deploy instructions now refrain from running the precommit formatter, even though it shouldn't happen anymore.

`pnpm-lock.yaml` is ignored, it shouldn't be formatted.

The offending files on `main` are reformatted to comply with the check.

Example failing output: https://github.com/opendatateam/udata-front-kit/actions/runs/30650941687/job/91223737202?pr=1327

Thanks @streino for the detective work 🕵️

### DIFF
--- a/scripts/deploy.sh
+++ b/scripts/deploy.sh
@@ -203,13 +203,13 @@ cmd_prepare() {
 
     # Merge source into merge branch
     info "Merging $source_branch into $merge_branch..."
-    if ! git merge "origin/$source_branch" --no-edit; then
+    if ! git merge "origin/$source_branch" --no-edit --no-verify; then
       error "Merge conflicts detected!
 
 Please resolve conflicts manually:
   1. Edit conflicted files
   2. Stage resolved files: git add <files>
-  3. Complete merge: git commit
+  3. Complete merge: git commit --no-verify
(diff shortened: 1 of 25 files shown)


abulte:
@streino since I'm off till Wednesday, can you merge this ASAP? 🙏

scripts/deploy.sh
```diff
@@ -203,13 +203,13 @@ cmd_prepare() {
...
     info "Merging $source_branch into $merge_branch..."
-    if ! git merge "origin/$source_branch" --no-edit; then
+    if ! git merge "origin/$source_branch" --no-edit --no-verify; then
       error "Merge conflicts detected!
 
 Please resolve conflicts manually:
   1. Edit conflicted files
   2. Stage resolved files: git add <files>
-  3. Complete merge: git commit
+  3. Complete merge: git commit --no-verify
```
streino, line 212: That one's tricky, because it would be safer to verify the changes resolved manually.

streino (approved):

PR #1342, opened 2026-08-11
feat(topics): enable discussion notification

### LINKED ISSUE #263: Notification à l'ouverture d'une discussion sur une collection
Lorsque un user commente un bouquet, l'owner du bouquet ne reçoit pas de notification.
Est de possible, à l'image d'un commentaire sur un jdd sur dgfr, que le owner d'un bouquet  reçoive une notification par mail quant à la publication d'un commentaire sur un bouquet. SDans le cas où le bouquet est owned par une orga, alors tous les users de l'orga reçoivent la notif
lorsque un nouveau commentaire est fait dans la discussion, l'ensemble des personnes ayant aussi commenté recoivent une notification

abulte:
Fix https://github.com/ecolabdata/ecospheres/issues/263

**This feature is only enabled for ecologie's collections.** Github is pinging everyone for review because I put the default value for the config flag to false on every site.

This plugs into the udata feature introduced in https://github.com/opendatateam/udata/pull/3765: when configured on a given page, add an extra with the "local" url for udata to point back to it in its notification email.

The notification email will include a deeplink like `#discussion-xxx` that is now handled. It currently does not handle pagination (target discussion has to be on first page). I think it's acceptable for this use case (notification should be fresh enough, and the discussion volume on topics is very low).

⚠️ The feature can not be tested e2e on demo.data.gouv.fr since it doesn't send emails. We can only verify the correct extra is sent and the deeplink is handled.

### DIFF
--- a/configs/ecospheres/config.yaml
+++ b/configs/ecospheres/config.yaml
@@ -250,8 +250,8 @@ definitions: &definitions
 #         display: {bool}             # display discussions tab on detail page
 #         create: {bool}              # create discussions tab on detail page
 #         notify_external_url: {bool} # on discussion creation, send extras.notification.external_url (this site's page for the subject)
-#                                      # so udata notification emails link back here instead of the subject's canonical udata page.
-#                                      # requires the site's domain to be allow-listed on the udata side (DISCUSSION_ALLOWED_EXTERNAL_DOMAINS). defaults to false.
+#                                     # so udata notification emails link back here instead of the subject's canonical udata page.
+#                                     # requires the site's domain to be allow-listed on the udata side (DISCUSSION_ALLOWED_EXTERNAL_DOMAINS).
 #       datasets:
 #         display: {bool}             # display datasets tab on detail page
 #       reuses:
--- a/configs/ecospheres/config.yaml
+++ b/configs/ecospheres/config.yaml
@@ -249,6 +249,9 @@ definitions: &definitions
 #       discussions:
 #         display: {bool}             # display discussions tab on detail page
 #         create: {bool}              # create discussions tab on detail page
+#         notify_external_url: {bool} # on discussion creation, send extras.notification.external_url (this site's page for the subject)
+#                                      # so udata notification emails link back here instead of the subject's canonical udata page.
+#                                      # requires the site's domain to be allow-listed on the udata side (DISCUSSION_ALLOWED_EXTERNAL_DOMAINS). defaults to false.
 #       datasets:
 #         display: {bool}             # display datasets tab on detail page
 #       reuses:
@@ -298,6 +301,7 @@ pages:
       discussions:
         display: true
         create: true
+        notify_external_url: false
       datasets:
         display: true
       reuses:
@@ -417,6 +421,7 @@ pages:
       discussions:
         display: true
         create: true
+        notify_external_url: false
       datasets:
         display: true
       reuses:
@@ -565,6 +570,7 @@ pages:
       discussions:
         display: true
         create: true
+        notify_external_url: false
       datasets:
         display: true
     editable: false
@@ -605,6 +611,7 @@ pages:
       discussions:
         display: true
         create: true
+        notify_external_url: true
       datasets:
         display: true
       reuses:
--- a/configs/ecospheres/config.yaml
+++ b/configs/ecospheres/config.yaml
@@ -249,9 +249,9 @@ definitions: &definitions
 #       discussions:
 #         display: {bool}             # display discussions tab on detail page
 #         create: {bool}              # create discussions tab on detail page
-#         notify_external_url: {bool} # on discussion creation, send extras.notification.external_url (this site's page for the subject)
-#                                     # so udata notification emails link back here instead of the subject's canonical udata page.
-#                                     # requires the site's domain to be allow-listed on the udata side (DISCUSSION_ALLOWED_EXTERNAL_DOMAINS).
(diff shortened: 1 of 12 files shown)


configs/ecospheres/config.yaml
```diff
@@ -249,6 +249,9 @@ definitions: &definitions
 #       discussions:
 #         display: {bool}             # display discussions tab on detail page
 #         create: {bool}              # create discussions tab on detail page
+#         notify_external_url: {bool} # on discussion creation, send extras.notification.external_url (this site's page for the subject)
+#                                     # so udata notification emails link back here instead of the subject's canonical udata page.
+#                                     # requires the site's domain to be allow-listed on the udata side (DISCUSSION_ALLOWED_EXTERNAL_DOMAINS).
```
streino:
```suggestion
#         notify_external_url: {bool} # when true, email notifications will link back to this site's pages instead of the canonical udata pages.
#                                     # website.seo.canonical_url must be set and the site domain must be allowed in the udata config 
#                                     # (DISCUSSION_ALLOWED_EXTERNAL_DOMAINS).
```

configs/ecospheres/config.yaml
```diff
@@ -249,6 +249,9 @@ definitions: &definitions
 #       discussions:
 #         display: {bool}             # display discussions tab on detail page
 #         create: {bool}              # create discussions tab on detail page
+#         notify_external_url: {bool} # on discussion creation, send extras.notification.external_url (this site's page for the subject)
```
streino: Shouldn't this be a generic parameter? I can't think of a case where I'd want topic discussions to link back to our site but datasets/... link back to datagouv.
This might not be working right away, but if that parameter makes more sense as a global one, I'd put it there now.

streino: I don't like `external_url` because 1) it speaks from the POV of udata instead of the vertical and 2) I'd expect to fill a URL instead of a bool.

Something like `notify_url_override` or `notify_force_site_domain` might be better?

src/components/DiscussionsList.vue
```diff
@@ -73,6 +73,20 @@ const pages = computed(() => {
 
 const allowDiscussionCreation = pageConf.resources_tabs.discussions.create
 
+// If configured, construct an external notification url from canonical_url for udata to use
+const getNotificationExternalUrl = (): string | null => {
+  if (!pageConf.resources_tabs.discussions.notify_external_url) return null
+  const canonicalUrl = config.website.seo?.canonical_url
```
streino, line 79: Ideally we should move `canonical_url` to something more generic than `seo`.

src/views/topics/TopicDetailView.vue
```diff
@@ -212,31 +215,55 @@ useMeta({
   noIndex: () => topic.value?.private
 })
 
-// Handle factor deeplinks: #factor-{id} switches to Données tab and scrolls to factor
+// Wait until `ready()` is true (e.g. a child component is mounted and its
+// data has loaded), then run `action()` once.
+const runWhenReady = (ready: () => boolean, action: () => void) => {
+  let done = false
+  const stop = watchEffect(() => {
```
streino, line 225:
```suggestion
  let stop: () => void
  stop = watchEffect(() => {
```
I had Claude explain `runWhenReady` (too lazy to read the `watchEffect` doc :/) and it suggested this fix:
> watchEffect invokes its callback synchronously, once, immediately, as part of the call to watchEffect(...) itself — before that call has even returned a value to be assigned to stop. That means:
> - If ready() is already true on that very first synchronous invocation, the callback tries to call stop() — but the const stop = ... assignment hasn't completed yet. Since stop is declared with const (block-scoped, in the temporal dead zone until initialized), this throws a ReferenceError: Cannot access 'stop' before initialization.

configs/ecospheres/config.yaml
abulte: Well that's what we're aiming for today: Topic discussions will link back to écologie, datasets won't. There are implications about hijacking datasets discussions that we need to plan around (if we even want to do it).

abulte: Yeah I couldn't find a better name, thanks for the ideas.

src/components/DiscussionsList.vue
abulte: Yeah, it's getting pretty critical. Creating an issue because it needs sync modifs on `udata-front-kit-seo` https://github.com/opendatateam/udata-front-kit/issues/1346.

src/views/topics/TopicDetailView.vue
abulte: My robot agrees.

configs/ecospheres/config.yaml
streino: Ok so no generic param for now

src/views/topics/TopicDetailView.vue
streino: :robot: 🤝 :robot:

streino: Suggestion not applied? Or github is acting weird on my end...

abulte: I went a different way, using a `let` would make linter complain. The new path shouldn't suffer from non-priming of `stop()`.

streino (approved):
https://github.com/opendatateam/udata-front-kit/pull/1342/changes#r3774291957 needs to be looked at, otherwise lgtm

src/views/topics/TopicDetailView.vue
streino: Right, sorry didn't pay attention :/

PR #1355, opened 2026-08-21
feat(ci): tag based docker build

abulte:
This introduces a new paradigm for future deployments: a normalised tag triggers a docker build and push.

Tag is `{site}-{env}-{YYYYMMDD}-{N}` like what the deploy script is already doing.

Two workflows:
- `build-push-image`: on a given tag (auto-triggered or passed via arg), build and push the docker image `{site}-{env}` for the given site with docker tags `{YYYYMMDD}-{N}` and `latest`.
- `create-deploy-release-via-tag`: meant to be used through GitHub UI, like `create-deploy-release`. Computes and creates the normalised tag, and calls `build-push-image` explicitly on that tag (GitHub won't trigger it automatically in that case).

The "real" deploy step still needs to be plugged-in. The new deploy infra is capable of auto-pulling `latest`, this should work right now. `deploy.sh` will need to be modified for the new workflow, either by relying on the latest trick or by triggering a deploy explicitly. This can be done later.

Re registry, the variable names and workflow are taken from https://github.com/datagouv/schema.data.gouv.fr/blob/main/.github/workflows/build-release.yml.

Actions view of a full chain:

Capture d’écran 2026-08-21 à 12 19 57

Command to test the UI workflow while not merged on main (tags `ecospheres-demo` branch and uses the `test` env):

```
gh workflow run create-deploy-release-via-tag.yml --repo opendatateam/udata-front-kit --ref feat/ci/build-image -f site=ecospheres -f environment=test -f ref=ecospheres-demo
```

Example build: https://github.com/opendatateam/udata-front-kit/actions/runs/32471960311

### DIFF
--- a/.github/workflows/build-push-image.yml
+++ b/.github/workflows/build-push-image.yml
@@ -79,6 +79,13 @@ jobs:
           echo "environment=$ENVIRONMENT" >> "$GITHUB_OUTPUT"
           echo "version=$VERSION" >> "$GITHUB_OUTPUT"
 
+      - name: Log parsed values
+        run: |
+          echo "matched: ${{ steps.parse.outputs.matched }}"
+          echo "site: ${{ steps.parse.outputs.site }}"
+          echo "environment: ${{ steps.parse.outputs.environment }}"
+          echo "version: ${{ steps.parse.outputs.version }}"
+
   build:
     needs: parse
     if: needs.parse.outputs.matched == 'true'
--- a/.github/workflows/build-push-image.yml
+++ b/.github/workflows/build-push-image.yml
@@ -1,17 +1,14 @@
+# Takes a tag like `<site>-<environment>-<YYYYMMDD>-<N>` and builds and pushes
+# `<site>-<environment>:<YYYYMMDD>-<N>` and `<site>-<environment>:latest` to registry
+
 name: Build and push site image
 run-name: Build and push ${{ github.event.inputs.tag || github.ref_name }}
 
 on:
   push:
     tags:
-      - 'ecospheres-*'
-      - 'meteo-france-*'
-      - 'logistique-*'
-      - 'accessibilite-*'
-      - 'defis-*'
-      - 'hackathon-*'
-      - 'simplifions-*'
-      - 'culture-*'
+      # Eligible tags are resolved later in `parse`, from SITES and ENVS vars
+      - '*'
   workflow_dispatch:
     inputs:
       tag:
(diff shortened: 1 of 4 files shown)


abulte:
@streino I could use a first review 🙏

.github/workflows/create-deploy-release-via-tag.yml
```diff
@@ -0,0 +1,111 @@
...
+        type: choice
+        options:
+          - ecospheres
+          - meteo-france
+          - logistique
+          - accessibilite
+          - defis
+          - hackathon
+          - simplifions
+          - culture
```
streino: Nitpick: sort lexically

.github/workflows/create-deploy-release-via-tag.yml
```diff
@@ -0,0 +1,111 @@
...
+          ref: ${{ steps.ref.outputs.ref }}
+
+      - name: Compute next tag
+        id: version
+        shell: bash
+        run: |
+          SITE="${{ inputs.site }}"
+          ENVIRONMENT="${{ inputs.environment }}"
+          TODAY=$(date +%Y%m%d)
+          PATTERN="${SITE}-${ENVIRONMENT}-${TODAY}-[0-9]+"
```
streino: Nitpick: could extract `${SITE}-${ENVIRONMENT}-${TODAY}` into a var, it's used a couple times in this step.

.github/workflows/build-push-image.yml
```diff
@@ -0,0 +1,92 @@
...
+      - name: Parse site, environment and version from tag
+        id: parse
+        shell: bash
+        env:
+          TAG: ${{ github.event.inputs.tag || github.ref_name }}
+          SITES: ${{ vars.SITES }}
+          ENVIRONMENTS: ${{ vars.ENVS }}
+          EVENT_NAME: ${{ github.event_name }}
+        run: |
+          if [[ "$TAG" =~ ^($SITES)-($ENVIRONMENTS)-([0-9]{8}-[0-9]+)$ ]]; then
```
streino, line 38: I didn't think `vars.SITES` could be used directly in a regex match. Really nice!

.github/workflows/build-push-image.yml
```diff
@@ -0,0 +1,92 @@
...
+      - name: Log parsed values
+        run: |
+          echo "matched: ${{ steps.parse.outputs.matched }}"
+          echo "site: ${{ steps.parse.outputs.site }}"
+          echo "environment: ${{ steps.parse.outputs.environment }}"
+          echo "version: ${{ steps.parse.outputs.version }}"
+
+  build:
+    needs: parse
+    if: needs.parse.outputs.matched == 'true'
```
streino, line 63: Nitpick: might want to add an explicit log for `matched != 'true'`, in addition to the "Log parsed values" above?

streino (commented):
Looks good!

The only thing I don't like too much is the `if ...matched == 'true'`, meaning we'll have to be careful about this if we add other steps. But I don't think we can do better :/

.github/workflows/build-push-image.yml
abulte: There's an explicit error in `workflow_dispatch` case. I think it's the only "risky" case, nobody should write the tags manually (`deploy.sh` will). So I'm fine with keeping this pretty silent; eg if we ever use non-deploy tags.

BTW there's an explicit log above:

```
echo "Tag '$TAG' does not match <site>-<environment>-<YYYYMMDD>-<N> (known sites: $SITES; known environments: $ENVIRONMENTS)"
```

streino (approved):

PR #1356, opened 2026-08-21
chore(indicators): remove org constraint on universe

abulte:
Will allow indicators from third-parties. Demo only for now.


streino (approved):
Should this go directly on the demo branch to avoid accidental merge to prod?

abulte:
@streino I should be able to handle it manually 🤞 I review the config file's diff for a deployment.

PR #1371, opened 2026-09-01
fix(matomo): use trackEvent from datagouv components

abulte:


abulte:
Fun fact: we inherit the events tracked by the components.

Capture d’écran 2026-09-01 à 08 43 14

streino (dismissed):

PR #1373, opened 2026-09-02
feat(datasets): enable new resources explorer

### LINKED ISSUE #1234: Intégrer le nouvel explorateur de ressources

abulte:
Fix https://github.com/ecolabdata/ecospheres/issues/1234

Enable the new resources explorer (opt-in via banner / persisted through localStorage) behind a configurable flag. Also enabled for indicators on ecologie.

A few things ~~to fix~~ fixed upstream:
- ~~fullscreen mode redirects to data.gouv.fr, but "Quitter le mode plein écran" does not redirect to the current site: add a prop for that~~ we now ship our own explore route (not that much work)
- Copy link button links to data.gouv.fr, not the current site: https://github.com/datagouv/cdata/pull/1215 https://github.com/datagouv/cdata/pull/1222
- Proper anchor for skiplinks feature: https://github.com/datagouv/cdata/pull/1216

Know issues:
- For indicators, when existing the full screen, we go back to the first tab instead of the files tab. Fixing that means implementing a deep-link navigation through tabs, out of scope here.
- Explore fullscreen view kind of "blinks" because the components from the normal page are dropped from view. Could be made better with a dedicated html entry point for explore view, but out of scope here.

~~This relies on a dev version of `@datagouv/components-next` for now, to be released upstream.~~

Capture d’écran 2026-09-02 à 11 52 03

Capture d’écran 2026-09-02 à 11 52 09


cypress/e2e/custom/ecospheres/datasets/files.cy.ts
```diff
@@ -0,0 +1,62 @@
...
+    })
+
+    const dataset = datasetFactory.one()
+    const resource = resourceFactory.one({
+      overrides: { title: 'Test resource' }
+    })
+    cy.mockDatasetAndRelatedObjects(dataset, [resource])
+
+    cy.visit(`/datasets/${dataset.id}`)
+    cy.contains("Revenir sur l'ancienne navigation").should('be.visible')
```
streino, line 50:
```suggestion
```
Redundant. Previous test ensures `resources_explorer_enabled` works, so I wouldn't re-check here.

src/views/datasets/DatasetExploreView.vue
```diff
@@ -0,0 +1,97 @@
...
+    params: { item_id: dataset.value?.slug ?? itemId },
+    query: typeof resourceId === 'string' ? { resource_id: resourceId } : {}
+  }
+})
+
+const loadError = ref(false)
+onErrorCaptured(() => {
+  loadError.value = true
+  return false
+})
```
streino, line 57: Might want to put the same comment as https://github.com/opendatateam/udata-front-kit/pull/1373/changes#diff-6576e002577a5abd20d60e7f638b99eb525ba2b21269fee234cfe092270d1768R46?

src/views/datasets/DatasetExploreView.vue
```diff
@@ -0,0 +1,97 @@
...
+        :exit-to="exitTo"
+        no-results-image="/static/blank_state/file.svg"
+        :resource-external-url="resourceExternalUrl"
+      />
+      <template #fallback>
+        <div class="fr-py-4w" role="status">Chargement de l'explorateur…</div>
+      </template>
+    </Suspense>
+  </div>
+</template>
```
streino, line 97: Very similar to https://github.com/opendatateam/udata-front-kit/pull/1373/changes#diff-6576e002577a5abd20d60e7f638b99eb525ba2b21269fee234cfe092270d1768R54-R71. 
I'm not sure I understand why we need both? Refactorable? Or maybe it's just a name thing (since this one seems to be fullscreen-only)?

src/router/index.ts
```diff
@@ -18,6 +19,23 @@ const defaultRoutes: RouteRecordRaw[] = [
     },
     component: async () => await import('@/views/HomeView.vue')
   },
+  // fullscreen resource explorer, opt-in per site
```
streino, line 22: Fullscreen is treated as a page on its own, meaning if you switch to fullscreen, go back, then hit the browser "back" button, you go back to fullscreen. I would expect the fullscreen toggle to have no impact on history.
In comparison, the new/old explorer toggle does not affect history (as I would expect).

Same behavior on data.gouv, so this is more a general concern that might be worth raising with them?

abulte: It is indeed a page of its own, with an URL and everything. It just looks like a fullscreen feature, but it's not. Maybe I shouldn't call it that 😅 but not sure I have a better idea. The icon is a bit misleading too.

streino: No better idea either, it does do it all like fullscreen... "Pretend fullscreen"? :sweat_smile: Maybe we should just open in another windown/tab so it's clearer it's a different page? Anyway probably needs more discussion, and sync w/ datagouv.

abulte: Yeah, for now it's iso data.gouv.fr, so I'll close this.

src/views/datasets/DatasetExploreView.vue
abulte: Refactored.

PR #1376, opened 2026-09-07
feat: add object counts on tabs

### LINKED ISSUE #1227: Indiquer le nombre d'objets contenus dans les sous-onglets d'une fiche dataset, service et collection
Ces indications sont disponibles dans l'UI data.gouv.fr. Pour les contributeurs, elles renforcent un sentiment de dynamique. Pour les simples utilisateurs, elles invitent à explorer les onglets.

Screenshot data.gouv.fr : 

Image

abulte:
Fix https://github.com/ecolabdata/ecospheres/issues/1227

Done for:
- datasets
- indicators (including sources)
- dataservices
- topics / collections (not including activity)

Capture d’écran 2026-09-07 à 09 40 10

### DIFF
--- a/cypress/e2e/custom/ecospheres/datasets/detail.cy.ts
+++ b/cypress/e2e/custom/ecospheres/datasets/detail.cy.ts
@@ -1,19 +1,12 @@
-import type { DatasetV2 } from '@datagouv/components-next'
-import { dataserviceFactory } from 'cypress/support/factories/dataservices_factory'
 import { datasetFactory } from 'cypress/support/factories/datasets_factory'
 import { resourceFactory } from 'cypress/support/factories/resources_factory'
-import { reuseFactory } from 'cypress/support/factories/reuses_factory'
 
 describe('Dataset Page - Tab counts', () => {
-  let testDataset: DatasetV2
-
-  beforeEach(() => {
+  it('should show distinct counts on each tab', () => {
     cy.mockMatomo()
     cy.mockStaticDatagouv()
 
-    // distinct, non-zero values so a field mix-up (e.g. summing the wrong
-    // metric, or reading a list length instead of resources.total) fails loudly
-    testDataset = datasetFactory.one({
+    const testDataset = datasetFactory.one({
       overrides: {
         metrics: {
           discussions: 7,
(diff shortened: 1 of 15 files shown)


cypress/e2e/custom/ecospheres/topics/detail.cy.ts
```diff
@@ -0,0 +1,23 @@
...
+  it('should show distinct counts on the Données and Discussions tabs', () => {
+    const { testTopic, testFactors } = setupTopicWithExistingFactors({
+      factors: createTestFactors(3),
+      discussions: discussionFactory.many(5)
+    })
+
+    visitTopic(testTopic.slug)
+    cy.wait('@getElementsDataset')
+
+    cy.contains('button', `Données (${testFactors.length})`).should(
```
streino, line 18:
```suggestion
    cy.contains('button', `Données (3)`).should(
```
More consistent with the other tests?

src/views/topics/TopicDetailView.vue
```diff
@@ -100,39 +108,44 @@ const breadcrumbLinks = computed(() => {
...
       tabId: 'tab-datasets',
       panelId: 'tab-content-datasets'
     })
   }
   if (showDiscussions) {
-    tabs.push({
+    result.push({
       title: 'Discussions',
+      count: topic.value
+        ? (discussionStore.getDiscussionsForSubject(topic.value.id)?.total ?? 0)
```
streino, line 126: Not fond of this extra request. Might be worth adding `metrics` to the topic response at some point.

cypress/e2e/custom/ecospheres/datasets/detail.cy.ts
```diff
@@ -0,0 +1,33 @@
...
+
+    cy.mockDatasetAndRelatedObjects(testDataset, resourceFactory.many(4))
+
+    cy.visit(`/datasets/${testDataset.id}`)
+    cy.wait(`@get_datasets_${testDataset.id}`)
+
+    cy.contains('button', 'Fichiers (4)').should('be.visible')
+    cy.contains('button', 'Réutilisations et API (5)').should('be.visible')
+    cy.contains('button', 'Discussions (7)').should('be.visible')
+    cy.contains('button', 'Informations').find('sup').should('not.exist')
```
streino, line 31: Checking absence of `sup` seems a bit fragile.
Claude suggests `contains(..., /^Informations$/)` instead?

src/views/topics/TopicDetailView.vue
abulte: Would be cleaner indeed, but clearly future.

cypress/e2e/custom/ecospheres/datasets/detail.cy.ts
abulte: Suggestion doesn't work. I don't think relying on very specific markup inside the button is that fragile? I made it more specific/explicit with the class.

streino: It's the combination of specific markup + absence. So if we change markup, checking for absence will still pass the test even though we're not looking for the right stuff anymore.
But if the suggestion doesn't work :shrug: Alternatives seemed too convoluted to be worth it.

streino (approved):

